PR #21, opened 2013-05-30
add Thruk Server

dcec:


HenriWahl:
@dcec Can you check if python3 based Nagstamon 2.0 still connects to latest Thruk 2?

PR #81, opened 2014-05-12
Uncaught exception in Zabbix code.

nono-gdv:
Another uncaught exception in the Zabbix code. This one can be triggered by a network timeout (in my case) or anything which can make an HTTP request fail.


HenriWahl:
Hi,
during my porting to Qt5 and Python3 I now converted Zabbix.py too. Can you please test the qt branch from https://github.com/HenriWahl/Nagstamon/tree/qt ? You might need to install the Python modules PyQt5, Requests and BeautifulSoup4 which both should be available via your distributions' package repository.
Thanks of anything new regarding Zabbix functionality.

PR #88, opened 2014-06-26
support SysTray on MacOS X

wfaulk:
This is not perfect, but it basically works.  The biggest problem is that "activate" is apparently never thrown under MacOS.  There should be ways to deal with left-vs-right button (other apps do), but I don't know what they are.


HenriWahl:
Obsolete with Qt5.

PR #130, opened 2015-02-21
add zabbix 2.4 support

arkadiyvleonov:


HenriWahl:
Thanks!
Does this still work with older Zabbixens?

arkadiyvleonov:
yes, work
in new API when you requesting api_version you don't need to use "auth" parameters unlike older zabbix.
may be it's not a good realisation, but it works

arkadiyvleonov:
Sorry, I wrote not about what was the issue:
Initial value of "auth" attribute should be "None" in ZabbixAPI Class

HenriWahl:
At https://seafile.ifw-dresden.de/d/c49b34412f/ there are Windows binaries containing this 2.4 Zabbix support - just in case someone needs them.

PR #134, opened 2015-02-27
Fix: issue 126 - set SSL_CERT_DIR environment variable and download CA c...

lhost:
Here is the correct way of OpenSSL certificate validation. More details here:
http://blog.hostname.sk/2015/02/27/nagstamon-python-openssl-certificate-validation/


HenriWahl:
Thanks for your contribution, but there are some issues:
 * this will not work in Windows, which most Nagstamon users use
* I want to replace the whole data retrieval stuff either by the powerful Requests module anyway
Thus I would prefer not to merge.

PR #135, opened 2015-03-02
Zabbix native statuses support

arkadiyvleonov:
with this patch nagstamon can display zabbix native severities of triggers which not exists in nagios - Information, High, average


HenriWahl:
Thank you. I can check this deeper after porting to Qt.

arkadiyvleonov:
sure. it's not urgent

HenriWahl:
Done with https://github.com/HenriWahl/Nagstamon/pull/354

Issue #140, opened 2015-03-31
[feature request] add clock to fullscreen mode


petyadev:
Add clock to the fullscreen mode, if it's possible.

Thanks,
Peter

HenriWahl:
Funny idea... yes I will try to remember this.

PR #141, opened 2015-04-01
add localtime to markup

arkadiyvleonov:


arkadiyvleonov:
for #140

HenriWahl:
While porting to Qt I would not pull GTK-related code, but I take this as inspiration when the time has come. Thanks!

PR #159, opened 2015-09-14
Pluggable servers support

marcinn:
Hello.

I would like to use Nagstamon to read state of services fom uncommon monitoring tool. It would be nice to register a Server component outside of Nagstamon.

That's why I've changed explicite servers registration to iteration over defined entry points (using entry point group `nagstamon.servers`). That gives possibility to register custom server by an external package, which solve my needs.

Side effects of this pull requests:
  * `setuptools` used instead of `distutils`, because `distutils` does not support entry points
  * servers list order was changed (in combobox)
  * invalid entry point (missing class or other error) will **NOT** break Nagstamon startup (ImportError catched); bad plugin should not crash whole app, so error is just logged

### DIFF
--- a/Nagstamon/Nagstamon/Custom.py
+++ b/Nagstamon/Nagstamon/Custom.py
@@ -21,7 +21,12 @@
 columns and other stuff.
 Imported in GUI module.
 """
+
 from Nagstamon.Actions import register_server
+
+import logging
+log = logging.getLogger('nagstamon')
+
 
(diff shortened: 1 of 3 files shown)


marcinn:
Nah, I've made some trashy commits. It would be better to do complete diff & patch instead of PR merge. Or I'll make new PR from new branch.

HenriWahl:
Apparently not mergeable...

PR #160, opened 2015-09-16
Do not remember old unacknowledged events in Zabbix.

nono-gdv:
The Zabbix API has a withUnacknowledgedEvents flag which returns triggers with unacknowledged event sometime in their history. All we really care about is the current state of the trigger, not its whole history.

This has been tested with Zabbix 2.2; I don't have access to earlier versions.


HenriWahl:
Dos not look like a major change so I will merge it.

PR #161, opened 2015-09-30
fix disabled_notifications visibility

cventastic:
disabled_notifications were visible, even if you filtered them out.


HenriWahl:
Thank you.
I will have to modify your code partly to work with Python 3 as the next Nagstamon 2.0 was ported to Python 3, but this won't be much.

PR #163, opened 2015-10-17
Make service acknowledgement work in Zabbix.

nono-gdv:
Zabbix requires some additional work to find the actual name of a trigger for acknowledging service issues. I'm not sure this is the best way to do it, but it works for me.


HenriWahl:
Must be merged later into qt branch too.

PR #169, opened 2015-12-04
Added LinuxMint to target dists

jhenkins:
This is to allow the DEB to build on LinuxMint 17, which is basically the same as Ubuntu 14.04 LTS.


HenriWahl:
Let's see how package builds will happen in 2.0 after it changed for Windows and MacOSX, but at least good point to have Mint on the list.

Issue #198, opened 2016-02-19
Nagstamon-qt displaying two weird rectangles


gtozzi:
I am not sure if this is a bug or a wanted feature.
If this is a feature, I didn't find a way to disable it and couldn't understand what it means, so there probably is still a minor bug behind.

I have just noticed nagstamon paces two small green rectangles on top-left of my desktop. They are always on top. They are movable: when they are created: they look like one, because they are on top of each-other. They are always on the current desktop (or in every desktop). If I click them nothing happens.

Here is a screen-shoot:
![istantanea_2016-02-19_21-49-40](https://cloud.githubusercontent.com/assets/386394/13188609/b57dfd18-d752-11e5-92df-37272658c37e.png)

gtozzi:
Update: right click opens the tray menu, but the "Servers" tab content is empty.

HenriWahl:
I wonder why there are two of them... and what is your display setting? Showing a statusbar or a systray icon? In the first case there should be only one and in the latter none... strange.
Do you have any servers configured?
Did you run it twice accidently?

gtozzi:
Thank you for your answer.
I have realized they were linked to two different processes. I killed them, and they vanished.

For some reason, they didn't read my configuration file (with 3 servers and systray mode) and went to default settings: no servers and statusbar.

Probably these processes were started by some sort of automated session-restoring in my desktop environment.

So in the end, no bug, just my fault. But I'm wondering... shouldn't the nagstamon process check if it is already running when started? Maybe the answer is no, for some reason I can't get, or this could be an option.

HenriWahl:
Some people use different instances of Nagstamon with different config files, so I do not think it might be a good idea to check if it is already running. In the worst case such a behavior would not have solved your problem if only running one instance with empty defaults.
Would be interesting if your both extra instances got a yet config argument by accident or session management which they may have considered as config directory to use.

gtozzi:
You are right: there was a --session argument that were misunderstood.
Here is a fix for this: https://github.com/gtozzi/Nagstamon/commit/9d8820fe3f44f31a2fb3513523cbc89859e0f036

PR #199, opened 2016-02-20
Better command line parsing, show the window on the right desktop

### LINKED ISSUE #197: Nagstamon-qt opening the window in the wrong desktop
Hi again,

I am using nagstamon-qt in systray mode.

When I click the icon, sometimes it does not open any window. If I then start cycling all the virtual desktops, I find the window opened in a different one.

Couldn't still determine which criteria it uses to choose the desktop, but often it does not chose the right one.

gtozzi:
This fixes #197 and prevents #198  from happening.

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -35,8 +35,6 @@ from PyQt5.QtCore import *
 from PyQt5.QtSvg import *
 from PyQt5.QtMultimedia import *
 
-import ewmh
-
 from operator import methodcaller
 from collections import OrderedDict
 from copy import deepcopy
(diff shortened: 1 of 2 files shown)


Nagstamon/QUI/__init__.py
```diff
@@ -1210,6 +1210,16 @@ def show_window(self, event=None):
...
 
+                    # Using the EWMH protocol to move the window to the active desktop.
+                    if not platform.system() in ('Darwin', 'Windows'):
+                        winid = self.winId().__int__()
+                        deskid = self.ewmh.getCurrentDesktop()
+                        self.ewmh.setWmDesktop(winid, deskid)
+                        self.ewmh.display.flush()
+
+                    # makes the window manager switch to the desktop where this widget has appeared
+                    self.raise_()
```
HenriWahl: Is the desktop really changing here? I intended to open the popup at the desktop where the user clicks the systray icon.

gtozzi: Yes, this makes the pop-up appear at the desktop where the user clicks the systray icon.

HenriWahl: Cool. i will check if it still works outside of XFCE :-)

gtozzi:
It also works on my KDE notebook.

Nagstamon/QUI/__init__.py
gtozzi: It also works on my KDE notebook.

HenriWahl:
I had to apply your changes only if in systray mode because floating statusbar mode did now behave strangely - see https://github.com/HenriWahl/Nagstamon/commit/ed23d25f8a315316b007afdb013f6bf6556016ed. Can you check if it still works as expected?

gtozzi:
Sorry, my fault: I didn't realize that function is being called for other modes too.
Yes, my mod was meant only for systray mode.
Yes, it still works as expected for me.
;)

HenriWahl:
Fixed now.

PR #200, opened 2016-02-21
Add Zabbix 3.0 support

oN0:
Hello!

Due to changes in Zabbix 3.0 API - added small patch.
Reason: removed ExpandData parameter in method trigger.get - https://support.zabbix.com/browse/ZBXNEXT-2724

Thanks.

### DIFF
--- a/Nagstamon/Nagstamon/Server/Zabbix.py
+++ b/Nagstamon/Nagstamon/Server/Zabbix.py
@@ -245,7 +245,6 @@ class ZabbixServer(GenericServer):
                 else:
                     state = '%s=%s' % (service['items'][0]['key_'], service['items'][0]['lastvalue']) 
                 n = {
-                    'host': service['host'],
                     'service': service['description'],
                     'status': self.statemap.get(service['priority'], service['priority']),
                     # 1/1 attempt looks at least like there has been any attempt
--- a/Nagstamon/Nagstamon/Server/Zabbix.py
+++ b/Nagstamon/Nagstamon/Server/Zabbix.py
@@ -260,7 +260,7 @@ class ZabbixServer(GenericServer):
                     'triggerid': service['triggerid'],
                 }
 
-                if api_version > '3.0':
+                if api_version >= '3.0':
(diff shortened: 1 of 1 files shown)


Nagstamon/Nagstamon/Server/Zabbix.py
```diff
@@ -261,6 +261,11 @@ def _get_status(self):
                     'triggerid': service['triggerid'],
                 }
 
+                if api_version > '3.0':
```
HenriWahl: should it be ">" or ">="?

oN0: Hmm, good question - I tried to pre-save the same code style as above in the code.

As they switched to 3-digit API version notation (3.0.0 returns now) both ">" and ">=" works, but if we will change version to "3.0.0" in code we should use only ">=".

Honestly, I'm not sure which one is more correct.

HenriWahl: So ">=" seems more correct - also better understandable for readers of thid code later if version has to be at least 3.0, regardless of the upcoming minor versions.

oN0: Yes, ">=" looks more reasonable.

Added.

HenriWahl:
Can you port thisth the master tree too? Otherwise I will put your changes into https://github.com/HenriWahl/Nagstamon/blob/master/Nagstamon/Servers/Zabbix.py

HenriWahl:
I added this to master tree. Can you check if the master versions runs in your setup too?

oN0:
I've checked the master version - you also need to merge this commit [e9df689](https://github.com/oN0/Nagstamon/commit/e9df689a0ffd11d25a9e7075acc65f7d7d28c16c) to have it working.

HenriWahl:
Applied, commited and pushed.

waardd:
Hello,

Im testing nagstamon with zabbix 2.2 and 2.4

I found that the information events are shown as unknown....
Also if i check the filter for maintenance they still show up in the monitor. 

Still testing... I'm really glad with this tool! Keep up the good work.
If i find more i will let you know.

Issue #192, opened 2016-02-15
v2.0 Acknowledgment fails


sronline:
we observed that the acknowledgment sometimes fails. The reason is that nagstamon uses the "display_name" instead of "service_description":

> define service {
>          #NAGIOSQL_CONFIG_NAME           nrpe_sesam_DATEN3_Musik
>          host_name                       BUP-A
>          service_description             NRPE_SESAM_DATEN3_MUSIK
>          display_name                    SESAM_DATEN3_MUSIK
>          check_command                   check_nrpe!check_sesam_DATEN3_Musik
>          max_check_attempts              3
>          check_interval                  180
>          retry_interval                  10
>          check_period                    24x7
>          notification_interval           600
>          notification_period             24x7
>          notification_options            w,u,c,r
>          notifications_enabled           1
>          contact_groups                  admins
>          failure_prediction_enabled      1
>          register                        1
>  }

![nagstamon_problem3](https://cloud.githubusercontent.com/assets/17158346/13044757/39795d28-d3d0-11e5-94c7-7d3b860c34fd.PNG)

Using the display_name leads to a "Not authorized" error:
![nagstamon_problem1](https://cloud.githubusercontent.com/assets/17158346/13044695/cffb2a7a-d3cf-11e5-9f84-e72bb65d242b.PNG)

HenriWahl:
This is Icinga, right?
If it fails sometimes I assume it does when display_name and service_description differ?

sronline:
Icinga 1.8.4. Yes it fails when display_name and service_description differ

HenriWahl:
Does this occur with other actions like "set downtime" too?

HenriWahl:
Done with latest commit - please check if acknowledges now. The thing about setting downtime would be of interest too.

minibbjd:
Tested with IcingaWeb2 and both acknowledge and downtime work fine for me.

yasa1987:
This bug correction introduced another bug when using version 2.0.1 and Icinga 1.7.1 (currently in the Debian stable repository) :
`ERROR: 2017-05-29 11:01:14.586506 Netmon Traceback (most recent call last):
  File "C:\Nagstamon\Nagstamon\Servers\Icinga.py", line 256, in _get_status_JSON
KeyError: 'host_name'`

This is line 256 :
`self.new_hosts[host_name].real_name = s['host_name']`
20 lines upper 's' is always tested against 'host_name' key existence but not this time. 
[...]

PR #202, opened 2016-02-27
Initial implementation for IcingaWeb2 server class

minibbjd:
The IcingaWeb2 REST API is not (fully) implemented yet, so currently this implementation is limited to "view only". Once https://dev.icinga.org/issues/9606 and/or https://dev.icinga.org/issues/7300 get implemented, the action part(schedule downtime, acknowledge, etc.) can be implemented in this class.

Most changes are in the new server class file Nagstamon/Servers/IcingaWeb2.py.

I've also changed  Nagstamon/QUI/__init__.py so that it only the context menu elements are shown that are set in server.MENU_ACTIONS. This might need checking if the variable is set correctly in all other server classes.

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -2745,25 +2745,33 @@ class TableWidget(QTableWidget):
             action_edit_actions = QAction('Edit actions...', self)
             action_edit_actions.triggered.connect(self.action_edit_actions)
 
-            action_monitor = QAction('Monitor', self)
-            action_monitor.triggered.connect(self.action_monitor)
+            if 'Monitor' in self.server.MENU_ACTIONS:
+                action_monitor = QAction('Monitor', self)
+                action_monitor.triggered.connect(self.action_monitor)
 
-            action_recheck = QAction('Recheck', self)
-            action_recheck.triggered.connect(self.action_recheck)
+            if 'Recheck' in self.server.MENU_ACTIONS:
+                action_recheck = QAction('Recheck', self)
+                action_recheck.triggered.connect(self.action_recheck)
 
-            action_acknowledge = QAction('Acknowledge', self)
-            action_acknowledge.triggered.connect(self.action_acknowledge)
+            if 'Acknowledge' in self.server.MENU_ACTIONS:
+                action_acknowledge = QAction('Acknowledge', self)
+                action_acknowledge.triggered.connect(self.action_acknowledge)
 
-            action_downtime = QAction('Downtime', self)
-            action_downtime.triggered.connect(self.action_downtime)
+            if 'Downtime' in self.server.MENU_ACTIONS:
+                action_downtime = QAction('Downtime', self)
+                action_downtime.triggered.connect(self.action_downtime)
 
             # put actions into menu after separator
             self.action_menu.addAction(action_edit_actions)
             self.action_menu.addSeparator()
-            self.action_menu.addAction(action_monitor)
-            self.action_menu.addAction(action_recheck)
-            self.action_menu.addAction(action_acknowledge)
-            self.action_menu.addAction(action_downtime)
+            if 'Monitor' in self.server.MENU_ACTIONS:
+                self.action_menu.addAction(action_monitor)
+            if 'Recheck' in self.server.MENU_ACTIONS:
+                self.action_menu.addAction(action_recheck)
+            if 'Acknowledge' in self.server.MENU_ACTIONS:
+                self.action_menu.addAction(action_acknowledge)
+            if 'Downtime' in self.server.MENU_ACTIONS:
+                self.action_menu.addAction(action_downtime)
 
             # not all servers allow to submit fake check results
             if 'Submit check result' in self.server.MENU_ACTIONS:
--- a/Nagstamon/Servers/IcingaWeb2.py
+++ b/Nagstamon/Servers/IcingaWeb2.py
@@ -24,10 +24,10 @@
 #
 # Status/TODOs:
 #
-# * The IcingaWeb2 REST API is not (fully) implemented yet, so currently this implementation is
-#   limited to "view only". Once https://dev.icinga.org/issues/9606 and/or https://dev.icinga.org/issues/7300
-#   get implemented, the action part (schedule downtime, acknowledge, etc.) can be implemented in
-#   this class.
+# * The IcingaWeb2 API is not implemented yet, so currently this implementation uses
+#   two HTTP requests per action. The first fetches the HTML, then the form data is extracted and
+#   then a second HTTP POST request is made which actually executed the action.
+#   Once IcingaWeb2 has an API, it's probably the better choice.
 
 
 from Nagstamon.Servers.Generic import GenericServer
@@ -55,9 +55,11 @@ class IcingaWeb2Server(GenericServer):
         object of Incinga server
     """
     TYPE = u'IcingaWeb2'
-    MENU_ACTIONS = ['Monitor']
+    MENU_ACTIONS = ['Monitor','Recheck','Acknowledge','Submit check result', 'Downtime']
     STATES_MAPPING = {"hosts" : {0 : "UP", 1 : "DOWN", 2 : "UNREACHABLE"},\
                      "services" : {0 : "OK", 1 : "WARNING",  2 : "CRITICAL", 3 : "UNKNOWN"}}
+    STATES_MAPPING_REV = {"hosts" : { "UP":0, "DOWN":1, "UNREACHABLE":2},\
+                     "services" : {"OK":0, "WARNING":1,  "CRITICAL":2, "UNKNOWN":3}}
     BROWSER_URLS = { "monitor": "$MONITOR-CGI$/dashboard",\
                     "hosts": "$MONITOR-CGI$/monitoring/list/hosts",\
                     "services": "$MONITOR-CGI$/monitoring/list/services",\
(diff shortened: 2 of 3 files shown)


HenriWahl:
Did you already check http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/icinga2-api#icinga2-api ?

minibbjd:
This is NOT a server class for the Icinga2 API, but for IcingaWeb2, just like the current Icinga server class is for the icinga-classic-ui.

A server class for the Icinga 2 API might be useful for some, too, but imho for most users using Nagstamon against IcingaWeb2 will be preferred. IcingaWeb2 has a pretty complete implementation for authentication (e.g. LDAP) and authorization (which users see which objects) and since most people will use these features for the IcingaWeb2 web interface anyway, it makes no sense to use the Icinga 2 API for Nagstamon.

HenriWahl:
Isn't Icinga2 the underlaying foundation of Icinga2web? Shouldn't Icinga2 API calls influence Icinga2web?

minibbjd:
IcingaWeb2 is written in php and has a defined, but limited API that can be accessed via the URL. I don't know how IcingaWeb2 processes API commands to Icinga2 - yes, maybe/likely through the Icinga2 API. 

Anyway, for this server class, we can only use the API that IcingaWeb2 (currently) offers. 

Accessing the Icinga2 API directly would need a setup with a different URL, username, password and port and fetching info and executing commands would need a different syntax and this would follow a different authentication/authorization scheme that IcingaWeb2. 

As such these are two separate server classes in Nagstamon.

HenriWahl:
I did not look into any Icinga2* API so far, I just saw the comments at https://mobile.twitter.com/Nagstamon/status/703579492946673664 which point me to Icinga2.

minibbjd:
Argh... a 3rd place to discuss.. :-)

HenriWahl:
Gives room for every opinion... :-) but least now I get your point about the differences of both ways to talk to Icinga. The worst case will be reverse engineering the POST requests or Ajax calls of the webinterface.

minibbjd:
My understanding is that there is no API at all in IcingaWeb2. There is an additonal module available at https://github.com/Thomas-Gelf/icingaweb2-module-deployment which only offers to schedule a downtime and which does not use the IcingaWeb2 authentication.

I've found a way to execute the needed command by analyzing the form data and making a second POST request to the same url with the necessary cgi_data. Works for recheck and acknowledge now and I will add 'Submit check result' and 'Downtime' soon.

HenriWahl:
This is how I would do it too.

minibbjd:
Can I expect the from start_time and end_time parameters in

_set_downtime(self, host, service, author, comment, fixed, start_time, end_time, hours, minutes):

to have a specific format? Currently the edit field is free text, but IcingaWeb2 needs something like "2016-02-28T20:03:55". I'm handling "n/a" now, but actually a datetime variable being passed, which might also be None seems to make more sense.

HenriWahl:
What is the edit field?

HenriWahl:
Sorry now I got it.
Every monitor server supported has some different format here - so it is no wonder that this applies here too. For other monitors like e.g. Nagios Nagstamon take the times from the parsed HTML at https://github.com/HenriWahl/Nagstamon/blob/master/Nagstamon/Servers/Generic.py#L458 .

HenriWahl:
Would you mind to call this Icinga2? Or do you expect some version working with icinga2 API?

MarcusCaepio:
I would call it Icinga2, because it uses the Icinga2 API,
How Can I test the new Nagstamon Version`?

HenriWahl:
Just checkout https://github.com/HenriWahl/Nagstamon.git.
If you need binaries the build.py script will produce them fitting to the platform where it runs on. You might need to meet some requirements which are listed on https://nagstamon.ifw-dresden.de/nagstamon-20/

MarcusCaepio:
Die Dev Umgebung herzustellen übersteigt leider meine Programmierkenntnisse..... 
Wäre es möglich, die fertige .exe auf die Nagsta Seite zu laden?

HenriWahl:
Eigentlich brauchst Du nur cx-freeze und InnoSetup. Ich versuche heute noch neue Binaries hochzuladen.

minibbjd:
> Would you mind to call this Icinga2? Or do you expect some version working with icinga2 API?

It's very likely that an Icinga2 API server class will follow, but I guess the IcingaWeb2 server class is what most people want. I'm not sure which naming is better...

Approach 1: Icinga2 (this one) and Icinga2API
Approach 2: IcingaWeb2 (this one) and Icinga2

minibbjd:
> I would call it Icinga2, because it uses the Icinga2 API,...

It does not use the Icinga2 API. It uses HTTP requests sent to the IcingaWeb2 web server.

HenriWahl:
Most common usage will be the IcingaWeb2 interface due to the included authentication and usermanagament as you outlined before. Thus I would assume this default and call it Icinga2. If one day someone wants to add an API version (which I consider unrealistic due to your working solution) it might get an APIesque name.

minibbjd:
Fine with me. Should I just change 

TYPE = u'IcingaWeb2'

to

TYPE = u'Icinga2'

or should the class and file name also be changed?

HenriWahl:
I already changed all occurrences of IcingaWeb2.

HenriWahl:
I took Icinga Web 2.2 and even login is not working apparently. Which Icinga version did you use?

HenriWahl:
Another remark: do we need to know the server version at all? This was necessary with Icinga1 to distinguish between pure HTML to be parsed and requesting JSON. Here it should be only JSON, right?

MarcusCaepio:
I have the current release, which came yesterday.
Version 2.2.0

minibbjd:
I'm still on IcingaWeb 2.1.2. I'll update tomorrow and check.

Yes, it's aways json, but sooner or later different versions if IcingaWeb2 will need different handling, so I left the implementation.

HenriWahl:
Ok, but it does not work with 2.2. I should create a 2.1 server for testing.

HenriWahl:
Problem seems to be getting a session - maybe login changed in 2.2?

minibbjd:
Seems like basic auth is broken in 2.2.

I've opened: https://dev.icinga.org/issues/11274

HenriWahl:
There is a login form on the Icinga start page. Is this different to Icinga 2.1?

minibbjd:
In 2.1, when you used basic authentication, like in my example in the issue I've opened, you were not redirected to the login page. The login page itself was also available in 2.1.

HenriWahl:
OK, just like with Check_MK. At the moment you only support basic auth?

HenriWahl:
Or simply always use auth page - are both ways of authentication valid and can be used in parallel?

minibbjd:
I assume this is a bug in 2.2. If not and devs consider basic auth to be insecure and desupported it, we would need to implement cookie auth through the login page. Let's wait for feedback in the issue I created.

HenriWahl:
In the meantime I implement cookie auth as this works well with other monitor types and is no rocket science.

minibbjd:
Not a bug: https://dev.icinga.org/projects/icingaweb2/repository/revisions/a464e74aa42142de039600f9630f1a8ce8c40218/diff/library/Icinga/Authentication/Auth.php

Basic auth only allowed when using IcingaWeb2 API calls (which are far from being usable).

HenriWahl:
Why oh why... give me some minutes :-)

HenriWahl:
Added login auth in last commit ac02140c9455d80565234ff19cc5d047640ce580. Apparently works, but now I get a new error:

  File "/home/henri/Seafile/Nagstamon/Nagstamon/Servers/Icinga2.py", line 188, in _get_status_JSON
    self.new_hosts[host_name].status = self.STATES_MAPPING['hosts'][h['host_state']]
KeyError: '1'

Can you check this with your Icinga 2.1?

minibbjd:
Strange... Works for me against 2.2.0. 

The error message looks like h['host_state'] is a string. For me it's an int, so the lookup in self.STATES_MAPPING['hosts'] works. Can you put a breakpoint there are check the types in h?

HenriWahl:
Got it with e2f450bdece88062739c1485ceef06d57a6beb18.
STATE_MAPPING used integers but JSON broght strings
Does the latest commit e2f450bdece88062739c1485ceef06d57a6beb18 still work for you?

HenriWahl:
If not the returned JSON values have to be converted then.

HenriWahl:
Something seems strange: here all hosts and services are acknowledged, flapping and in downtime in Nagstamon which they are not on webinterface.

HenriWahl:
Might have to do with misunderstood JSON values too.

minibbjd:
No I get the error:

  File "C:/Users/Marcus.Moennig/Nagstamon\Nagstamon\Servers\Icinga2.py", line 187, in _get_status_JSON
    self.new_hosts[host_name].status = self.STATES_MAPPING['hosts'][h['host_state']]
KeyError: 1

I guess we have to typecast everything.

HenriWahl:
You get integers and no strings? Strange. Icinga version dependent? 

I meanwhile fixed the useless extra icons - coming from strings in JSON too. This fix will be obsolete if all JSON values from Icinga will be integers where useful.

HenriWahl:
From https://docs.python.org/3.4/library/json.html:

json.loads(s, encoding=None, cls=None, object_hook=None, parse_float=None, parse_int=None, parse_constant=None, object_pairs_hook=None, **kw)¶

    Deserialize s (a str instance containing a JSON document) to a Python object using this conversion table.
    The other arguments have the same meaning as in load(), except encoding which is ignored and deprecated.
    If the data being deserialized is not a valid JSON document, a ValueError will be raised.

Looks like this can be done easily.

minibbjd:
Probably depends on the version of the python json library (I'm using python 3.4.3). 

Your last commit doesn't work for me:

File "C:/Users/Marcus.Moennig/Nagstamon\Nagstamon\Servers\Icinga2.py", line 187, in _get_status_JSON
    self.new_hosts[host_name].status = self.STATES_MAPPING['hosts'][h['host_state']]
KeyError: 1

I'd suggest to typecast to int here, too and change the MAPPING definitions back to int, too.

minibbjd:
One more thing: IcingaWeb2 and Icinga2 have a service/host attribute "service_problem" and "host_problem" (click on "Problems" in the web interface). It would be nice to be able to filter for these in Nagstamon, too. The benefit is that also services/hosts that have a dependency on other hosts/services are filtered out, when the parent host/service is down. 

I didn't implement it, because it would be a filter in the filter settings dialog that would only work with Icinga 2. Wasn't sure if cluttering up the interface for just one server implementation would be the right way to go.

HenriWahl:
OK I'll do the typecasting.
For the filter let's, see, I would like to make the basics of Icinga work first. There was somewhen a feature request about server-specific filters, maybe this would be a good place then.
Did you try to trigger some action when logged in? Did not work for me :-(

HenriWahl:
Did the typecasting. Does it work at your setup?

minibbjd:
Typecasting works now.

Seems like one of my commits is lost. To fix downtime action for hosts change around line 422:

        if service=='':
            formtag=pagesoup.find('form',{'name':'IcingaModuleMonitoringFormsCommandObjectScheduleHostDowntimeCommandForm'})
        else:
            formtag=pagesoup.find('form',{'name':'IcingaModuleMonitoringFormsCommandObjectScheduleServiceDowntimeCommandForm'})

The other actions seem to work just fine for me.


https://dev.icinga.org/issues/11274 indicates that all that's actually needed is 

 self.session.headers['Referer'] = 'application/json'

to qualify as an API request and use basic auth.

HenriWahl:
I read there abaut the 'Accept' header... but well, so all the BeautifulSoup stuff for triggering actions is not necessary?

minibbjd:
No, this is still needed. The IcingaWeb2 just call it an API call when the http header "Accept: application/json" is set. There still is no working API, so we have to parse the form tags.

When you set 

self.session.headers['Accept'] = 'application/json'

you don't have to go through the login page and work with the cookie, but can just use basic auth like in my first implementation.

minibbjd:
Side question: For a special project at work I will need to write a server class that does no http requests, but parses a Nagios/icinga status.dat file (copied every 5 minutes from an Icinga instance in an isolated network subnet). Should I branch and request a pull or do you think this is too exotic to be included. :-)

HenriWahl:
OK. Too late now... cookies are now accepted. :-)

Your side project sounds pretty special - maybe just fork but do not request a pull, you seem to be the only user.

HenriWahl:
8bdc35fff2340e15a6a0943648d4ad80a53ee98d brings now display of start end times in downtime dialog

dnsmichi:
I'd suggest to name the backend "icingaweb2" to avoid confusion with "icinga2" providing a different api. Users might not be aware of the difference, and there certainly are differences between the two products. That terminology to distinguish between the core and web is commonly used in the Icinga ecosystem. It is also to avoid questions in the community support channels :)

HenriWahl:
OK. Are there plans to provide a fully supported Icinga2 API equivalent to the icingaWeb2 API regarding user authentication?

dnsmichi:
Icinga 2 allows you to specify the ApiUser configuration object. You may set permissions on the specific API endpoints. If you need a more granular level, you can also add filters for these permissions. For example, you'll only allow an acknowledgement action for this specific ApiUser, but only for all hosts and services in a defined hostgroup. Or matching a custom attribute. 
If you prefer to have more than one ApiUser with a defined filter set, go for it. If you don't like basic auth but want to use an SSL certificate for client CN auth, that's also possible.

For more details please check the documentation: http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/icinga2-api#icinga2-api-permissions

HenriWahl:
It is renamed now in commit 8fbbe360523f7b0bb288fc914cd9fd8b597b29d6.

dnsmichi:
Thanks a lot :)

HenriWahl:
Does the error as in https://github.com/HenriWahl/Nagstamon/issues/192 occur with IcingaWeb2 too?

minibbjd:
I'm sick, so can't test it right now, but in my implementation I am only using service and host names and not display_name. If you haven't changed it, it should work just fine.

HenriWahl:
But the concept of display_names exists in IcingaWeb2 too? If you do not use them I should deactivate the corresponding options in settings GUI.

minibbjd:
Yes, see: http://docs.icinga.org/icinga2/latest/doc/module/icinga2/toc#!/icinga2/latest/doc/module/icinga2/chapter/object-types#objecttype-service

diplay_name is not used. 

However, it would be perfect if both, display_name and service/host name would be fetched and stored, display_name would be used for displaying and the stored service/host name would be available when executing actions.

HenriWahl:
This is how I do it now with Icinga1.
Looks like display_name is just optional or is there any default, like parity with host or service name?
I will make my current implementation more robust and adapt it to IcingaWeb2.py.

HenriWahl:
I added display_name awareness to all IcingaWeb2 actions - maybe you could test with some real life setup if it still works and if the difference between real names and display names is recognized and treated accordingly?
Additionally I added the correct URLs for the 'Monitor' action in context menu, choosing the URLs which looked best. Can you as Icinga2 user have a look if they make sense? Thanks!

minibbjd:
IcingaWeb2 doesn't like spaces being replaced to + signs in URLs which is what urllib.parse.urlencode does. I've changed the two calls in IcingaWeb2.py open_monitor like this:

urllib.parse.urlencode({'host': self.hosts[host].real_name})**.replace('+', ' ')**)


urllib.parse.urlencode({'host': self.hosts[host].real_name,'service': self.hosts[host].services[service].real_name})**.replace('+', ' ')**)

Issue #189, opened 2016-02-12
v2.0 application can be started twice


sronline:
the application should check if it has already been started

HenriWahl:
Why?

There are users running several instances with different configuration.

gtozzi:
I am also interested on this: there also are users not willing to risk running several instances by mistake: when the session gets saved and restored, it could happen :)

Could this be an option? I could write the code for it using the lockfile (https://pythonhosted.org/lockfile/lockfile.html) module, but the hard part is adding the checkbox to the ui ;)

HenriWahl:
The GUI part is not that hard, I just imagine it somehow confusing if you have one config allowing multiple instances and one not...

gtozzi:
True. But i think then we can say nobody has a reason to run two instances over the same config. Am I wrong?

HenriWahl:
No, you are right. So maybe it is enough to check at start if there is an process with the same command line args?

gtozzi:
I don't think this is doable in a safe and portable way.
But it could place a lock on the config folder using the lockfile module or by creating a pid file inside it. If the lock fails, then there is another instance holding the lock for the same folder.

HenriWahl:
Doesn't this bring even more overhead? Maybe it would be enough to evaluate a --session argument and then check if already some Nagstamon runs? This would only happen on Linux/BSD anyway.

gtozzi:
What do you mean by overhead? Locking a folder or creating a 5-bytes file happens in the blink of an eye.
In my case there is not always a --session argument. It could also happen on Windows, if you have it both on autostart and in the register, by example, or if you start it manually while it is being autostarted on slow machines; i don't know about OSX.

HenriWahl:
Overhead means dependencies, checks and stuff. New error source like Nagstamon-crashed-but-PID-file-still-hanging-around-and-blocking-new-start. But you are right, quite possibly this could happen on other OS too.
I found the psutil module which promises to let one access processes in a platform agnostic way without any PID files involved - might be worth a look too.

gtozzi:
The psutil module does not seem to do that, what method/function did you see?
There is a tendo.singleton module that aims to do that, but it only applies to any instance without option to discern between config paths.
Checking the command line path is not an option because some OS/variants mess up with it and, anyway, what if i invoke one instance with -psn and the other bymanually passing ~/.nagstamon as config folder? They both use the same config but have different command lines.
[...]

PR #208, opened 2016-03-01
More readable code on argument parser

gtozzi:
This should do the same, but with more readability (and arguably in a more "pythonic" way).


Nagstamon/Config.py
```diff
@@ -179,20 +179,20 @@ def __init__(self):
 
         # Parse the command line
         parser = argparse.ArgumentParser(description='Nagios status monitor for your desktop')
-        parser.add_argument('cfgpath', nargs='*', help='Path for configuration folder')
+        parser.add_argument('cfgpath', nargs='?', help='Path for configuration folder')
```
HenriWahl: My last finding was nargs='?' does not work with build script, only nargs='*'

Nagstamon/Config.py
```diff
@@ -179,20 +179,20 @@ def __init__(self):
...
         # if sys.argv is larger than 1
-        if args.cfgpath:
-            # MacOSX related -psn argument by launchd
-            if '-psn' in args.cfgpath:
-                # new configdir approach
-                self.configdir = os.path.expanduser('~') + os.sep + '.nagstamon'
-            else:
-                # allow to give a config file
-                self.configdir = args.cfgpath[0]
+        if args.psn:
```
HenriWahl: what if there is no arg '-psn'? I assume this check would fail on anything not OSX

Nagstamon/Config.py
gtozzi: You used nargs='*' and then config[0] as a workaround for not adding a '-psn' argument. This is supposed to work the same. The build-script related part should be only the known_args() call.

Nagstamon/Config.py
gtozzi: Yes, if there is no -psn this is False. Isn't how it is supposed to work?

HenriWahl: No it will crash if args does not have the property .psn - this is why I checked if it has it.

gtozzi: It does not happen to me. Can you please paste the trace?

HenriWahl: At least I had it yesterday with another argument. I will check this - right now I try to fix the build script for Debian which I already finished.

gtozzi: I am sorry I've got confused. I wrote this code this morning, how could it crash yesterday?
Anyway, take your time ;)

Thank you for your work.

HenriWahl: I tried yesterday this args.something where it crashed if there was no .something given as --something at the commandline.

HenriWahl: But I should not do three things at the same time... :-)

gtozzi: Please just give it a try when you'll have time ;)

HenriWahl: Promised!

gtozzi:
I've added #189 to this pull request ;)

HenriWahl:
Works great, thanks! I will just look forward to add GUI option and consider https://github.com/HenriWahl/Nagstamon/issues/189#issuecomment-191244586.

gtozzi:
Thank to you ;)
About #189 comment, I have already included the username in the files, so it should also address that.
Probably the GUI options is really not needed anymore. Unless somebody really wants to run two instances on the same machine from the same user on the same config :P

HenriWahl:
But different users will overwrite the same file, right?

gtozzi:
Yes, if multiple users are using the same config folder at the same time, the lock file gets overwritten.
It's not the same for the config itself?

HenriWahl:
Errm.. yes. :-)

PR #214, opened 2016-03-23
Plus signs in monitoring URLs and find version number with all languages

minibbjd:
Fixes getting version number when language is not English or  German.
Fixes monitoring URLs which should not have plus signs.


HenriWahl:
Thanks for the fixes!

PR #221, opened 2016-04-13
Check_MK Multisite: Added server option to enforce AuthUser

LarsMichelsen:
When a user which is permitted to see all hosts/services in Check_MK
Multisite is used from Nagstamon it displays the problems of all
hosts/services.

Some admin users only want to see "their" hosts. These are the hosts
they are assigned with as contact.

Since Check_MK 1.4.0i1 the new option to "show only permitted hosts"
is supported by Check_MK.

Previous Check_MK versions had a user profile option named "Visibility of
Hosts/Services (Webservice)" to configure this. But this option has
been deprecated and will be removed in future releases. If you used this
option before you are adviced to disable it and enable the new option
in Nagstamon.

### DIFF
--- a/Nagstamon/Config.py
+++ b/Nagstamon/Config.py
@@ -731,6 +731,11 @@ class Server(object):
         self.use_display_name_host = False
         self.use_display_name_service = False
 
+        # Check_MK Multisite
+        # Force Check_MK livestatus code to set AuthUser header for users who
+        # are permitted to see all objects.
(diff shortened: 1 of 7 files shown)


golimer:
KeyError: 'Host_notification_enabled'

any idea hot to fix?

HenriWahl:
Fixed with commit ce2dd6771c9f547ceb35a1ae04017b0a3495ed0c

PR #226, opened 2016-04-18
quote string pushed to browser

theundefined:
quote string passed to browser

### DIFF
--- a/Nagstamon/Nagstamon/Actions.py
+++ b/Nagstamon/Nagstamon/Actions.py
@@ -657,7 +657,7 @@ class Action(threading.Thread):
                         }
             # mapping mapping
             for i in mapping:
-                string = string.replace(i, mapping[i])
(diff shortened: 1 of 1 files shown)


HenriWahl:
Just out of curiosity: is there any URL not working if being not quoted?

HenriWahl:
Just another remark: this branch is dead and just remaining for historical reasons. Please check the master branch if you want to contribute.

theundefined:
hm... version 1.0 is no longer supported? :(


problem is discovered when i created action to create form in jira to create ticket. It works ok, if in string from nagios there is no % character (for example 90% used ;). After escaping it - it works fine.

HenriWahl:
Please try with 2.0. Maybe your problem is gone already there.

HenriWahl:
@theundefined did you check with Nagstamon 2.0.1?

PR #244, opened 2016-05-12
op5Monitor: Service is in downtime if its host is

### LINKED ISSUE #237: OP5 and Nagstamon schedule downtime not correct
When scheduling a host in OP5 in downtime the services also are being put in downtime.
Unfortunally Nagstamon doest reconize this, so when the host goes down Nagstamon keeps notifying that the services are down. (All filters are set correctly) This issue is present in all versions of Nagstamon (also the stable release).

http://oi68.tinypic.com/msioty.jpg
http://oi68.tinypic.com/wvoevm.jpg

Running Nagstamon on Windows 10 against OP5 v7.2.0

chelmertz:
A service should be treated as "in scheduled downtime" either if
itself is marked as such, or the service's host is in scheduled downtime.

This fixes #237

Signed-off-by: Carl Helmertz <chelmertz@op5.com>

### DIFF
--- a/Nagstamon/Servers/op5Monitor.py
+++ b/Nagstamon/Servers/op5Monitor.py
@@ -92,6 +92,7 @@ class Op5MonitorServer(GenericServer):
     api_svc_col.append('host.name')
     api_svc_col.append('host.state')
     api_svc_col.append('host.active_checks_enabled')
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks!

PR #317, opened 2016-09-23
Icinga: Fix version detection when setting URLs

jplitza:
The previous version comparison was flawed, as it only compared strings and thus '1.10' < '1.7' was true. This commit uses the `distutils.version` package to properly compare versions.


HenriWahl:
Thanks! Might be a good idea to use in other parts of Nagstamon too.

PR #329, opened 2016-10-18
Centreon 2.8 beta, code refactoring

BenoitPoulet:
Centreon 2.8 beta
Code refactoring
Bug fixes

### DIFF
--- a/Nagstamon/Servers/Centreon.py
+++ b/Nagstamon/Servers/Centreon.py
@@ -36,14 +36,10 @@ class CentreonServer(GenericServer):
     # centreon generic web interface uses a sid which is needed to ask for news
     SID = None
 
-    # Centreon have different URL paths for XML
-    # will be checked by _get_xml_url() but default is xml/broker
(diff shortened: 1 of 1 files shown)


HenriWahl:
Cool! I will merge this after release of currently stable 2.0.1.

BenoitPoulet:
Ok, i understand.

It would be nice if you could build a beta version with this when Centreon 2.8 will be out.

HenriWahl:
Yes, I think about going to a numbering scheme like GIMP, having even numbers for stable and odd ones for unstable/testing releases. So your changes will first go into 2.1 after this bugfix release 2.0.1 is ready - which I think it almost is.

HenriWahl:
@BenoitPoulet I started 2.1 with your merges. What binaries do you need?

BenoitPoulet:
Don't need to hurry, i ask them and the 2.8 release will not be out before end of november.
I just need some user's feedback to be sure that everything is fine before you next stable version.

PR #339, opened 2016-11-29
Add Kerberos authentication support

### LINKED ISSUE #107: Support for kerberos authentification
Hi,

I'm an avid user of nagstamon, but recently we integrated kerberos for authentification on our shinken/check_mk install and now I can  only use the web interface, nagstamon doesn't seem to support kerberos. Any chance of getting this on nagstamon ?

sim0nx:
The following PR implements Kerberos authentication support and fixes #107 .
Would be great if you could merge it!

Thanks


HenriWahl:
Cool! Do you have any experience how does this work on other Linux distributions and OS?

sim0nx:
I personally use this on a Debian machine.
As the patch is not intrusive and basically only requires requests-kerberos (which is an officially supported extension see [0]) my guess is that it just works.

requests-kerberos either depends on the python kerberos or winkerberos module (see [1]). Which tells me that it is compatible with windows as well.
OSX is supported as well (the kerberos module is from Apple).

There are Debian/Ubuntu and Centos/Fedora/RH packages from what I know. Other distros surely package them as well.

So as long as requests-kerberos and kerberos modules are installed it just works.

[0]: https://github.com/requests/requests-kerberos
[1]: https://github.com/requests/requests-kerberos/blob/master/requests_kerberos/kerberos_.py

mprenditore:
I cannot test it now, but due to the small modifications in the code I think it cannot break anything.

HenriWahl:
Sounds good. If I find time I will test this on several OS.

HenriWahl:
By the way: as Active directory is based on Kerberos this is assumed to work on any Windows machine too? In theory not even should need a username to be given as the login is used?

mprenditore:
@HenriWahl Kerberus doesn't require a username because of how it's work... AFAIK it's a client that store credential system wide and communicate them to the server generating a token.

HenriWahl:
Yes, I mean, so the normally used username and password in Nagstamon are obsolete if Kerberos is used. Thus they might be omitted in GUI in this case.

sim0nx:
@HenriWahl yes this should work with a domain-joined windows as well. Can't test that though.
And I agree that if Kerberos is selected, the username and password could in fact be omitted.

Thanks for merging!

mprenditore:
I agree too about removing username and password if Kerberos is selected

PR #354, opened 2017-01-20
Zabbix Trigger severity compatibility and some code clean

mprenditore:
Based on the pull request of @arkadiyvleonov "Zabbix native statuses support #135" I added the Zabbix DEFAULT Trigger Severity compatibility
I know that in the future the Severity should be revisited but at least, now, we have this support.

I hope everyone enjoy it.


HenriWahl:
Sorry for the late reply - a lot of stuff to do at work, 
Is Nagstamon still running after all the cleaning? :-)

mprenditore:
Usually the code cleaning means pep8 compliance, removing trailing spaces, radding default "except Exception" or a proper exception instead of a basic "except" and indent code little bit better...
In that case I also fixed the "self.exec" that in some interpreter could give errors changing it to "self.exec_", that's because "exec" is a core function of Python but PyQt5 have also the same function in the class, but reading the documentation is possible to call it with "exec_"

Testing on Fedora 25 qith Qt5.7... After 20 days since the code cleaning no error or crash and the new implementation seems works very well ;)
I monitor 2 Zabbix servers and 2 Centreon servers and everything is working well... Do your test and let me know :)

HenriWahl:
Runs on my Fedora 25 too checking no Zabbix but Check_MK. Great!

mprenditore:
Now I'll work to the exception when a server could not be reached for some reason...

HenriWahl:
I have no time to investigate further but I have a funny phenomen here: if there are only warnings the systray icon turns GREEN instead of YELLOW. Do you have some similar effect there?

mprenditore:
I can confirm the bug... Thanks to pointed it out... All other colours seems to work... I'll investigate why warning doesn't work because I didn't touched it, it's the original one, maybe i touched something too much...
I should see it before tomorrow

HenriWahl:
Thanks for investigating. I assume the new number of possible state might cause some miscalculation somewhere.

mprenditore:
I'm thinking the same... I'm already checking it :)

mprenditore:
Fixed in last pull request...
I forgot one check about the worst state.
Thanks for finding it out :)

HenriWahl:
Merged already and tested - works! Thanks.
What a pity to have no green icon anymore... :-(

mprenditore:
I feel you... I have a bordeaux one :(

PR #382, opened 2017-04-23
Windowed mode

oernii:
This PR creates a new 'Windowed mode', which allows for a non-fullscreen resizable window with tray icon.


HenriWahl:
Cool - sounds like a very useful addition. I will check asap.

Nagstamon/QUI/settings_main.py
```diff
@@ -1347,7 +1349,7 @@ def retranslateUi(self, settings_main):
         self.label_colors.setText(_translate("settings_main", "Set colors for text and background of status information:"))
         self.states_groupbox.setTitle(_translate("settings_main", "Status:"))
         self.label_color_ok.setText(_translate("settings_main", " OK "))
-        self.label_color_information.setText(_translate("settings_main", " INFORMATION "))
+        self.label_color_information.setText(_translate("settings_main", " INFORMATIOn "))
```
mprenditore: Typo to fix on the last "n"

Nagstamon/QUI/settings_main.py
```diff
@@ -1265,27 +1267,27 @@ def retranslateUi(self, settings_main):
...
         self.input_checkbox_filter_all_critical_services.setText(_translate("settings_main", "All critical services"))
-        self.input_checkbox_filter_hosts_services_disabled_checks.setText(_translate("settings_main", "Hosts and services with disabled checks"))
-        self.input_checkbox_filter_all_flapping_hosts.setText(_translate("settings_main", "All flapping hosts"))
-        self.input_checkbox_filter_services_on_hosts_in_maintenance.setText(_translate("settings_main", "Services on hosts in maintenance"))
-        self.input_checkbox_filter_services_on_acknowledged_hosts.setText(_translate("settings_main", "Services on acknowledged hosts"))
         self.input_checkbox_filter_all_unknown_services.setText(_translate("settings_main", "All unknown services"))
-        self.input_checkbox_filter_all_high_services.setText(_translate("settings_main", "All high services"))
-        self.input_checkbox_filter_all_average_services.setText(_translate("settings_main", "All average services"))
+        self.input_checkbox_filter_all_high_services.setText(_translate("settings_main", "all high services"))
```
mprenditore: "all" should be "All" as all the above and below texts

FibreFoX:
@HenriWahl Any progress on this? This PR might be even working for this issue: https://github.com/HenriWahl/Nagstamon/issues/396

oernii:
Well, I cannot pinpoint a problem with multiple monitors, where the window reappears on the wrong one. Also I cannot seem to be able to remove the annoying flicker which occurs after refresh. Will look into in some more, hopefully this week.

HenriWahl:
Why should there be a window and a systray icon? What happens when clicking the systray icon? Shouldn't the window mode rather be derived from fullscreen mode?

FibreFoX:
Quite simple: multi-monitoring (especially if there are more than two monitors) . If you work on the main-system, you might want to have a blinking tray-icon there, and a quick view on a different monitor while having this open. As the systray-icon is only a quick way of opening the full list, there shouldn't happen anything in addition. Another thing to have both: checking if the application is still runing (happens to me, that I search for "background"-programs inside my icontray first).

HenriWahl:
So in certain moments there will be 2 windows showing the status, one running all the time and one popping up when clicking the systray icon?

Issue #184, opened 2016-02-06
Nagstamon 2.0 can't open status


gtozzi:
Hi there,

I'm testing the current GIT snapshot on KDE5.
After installing the required python modules it runs, sets up the tray icon, and connects to server.
But when I click the icon, the status window only "flashes" for a brief moment, then closes again.

HenriWahl:
Is there any output in a terminal?

gtozzi:
Apparently nothing relevant. I am using lastest Kubuntu with KDE 5.15.

This is the output at startup, but I get no extra output when i try to click the icon.
>$ env LANG=C ./nagstamon-qt.py
>QLayout: Attempting to add QLayout "" to StatusWindow "", which already has a layout
>QLayout: Attempting to add QLayout "" to StatusWindow "", which already has a layout
>QLayout: Attempting to add QLayout "" to StatusWindow "", which already has a layout
>GStreamer; Unable to pause - "file:///mnt/secfs/Nagstamon/Nagstamon/Nagstamon/resources/down.wav"
>Warning: "No decoder available for type 'audio/x-wav'."
>Error: "Your GStreamer installation is missing a plug-in."

If you can tell me where to start looking for, I can try to debug the source.

HenriWahl:
There is a debugging mode, available at first tab of settings.
Anyway I think I will try myself if I can reproduce this.
When I see the error message you got - maybe you could make sure that

> Warning:  "No decoder available for type 'audio/x-wav'."
> Error: "Your GStreamer installation is missing a plug-in.

wouldn't happen?

gtozzi:
I have disabled sound and enabled debug log.
The gstreamer error is gone, the debug log is empty.
I have debug messages straight into the terminal, but they only refer to:
> DEBUG: 2016-02-07 08:02:52.732645 servername FetchURL: ...

HenriWahl:
You use Kubuntu 15.10, right? Just to make sure to reproduce it correctly.

gtozzi:
Affirmative.

HenriWahl:
I have it running now in a VirtualBox VM:
- Kubuntu 15.10 64 bit
- displayed as systray icon

When I click it the first time after starting it I see a short flashing status window, every next click works flawlessly. You have the short flash every time?

gtozzi:
Yes.

HenriWahl:
[...]

PR #393, opened 2017-06-25
Increase flicker prevention delay to .5 secs

gtozzi:
Hi! :) #184 is back.

The problem happens in Linux/KDE when the window height is bigger than the
screen height. As soon as the window is shown, it gets on top of the tray bar
and an "enterEvent fires". Moments later (sometimes even .2 secs later), the
tray bar gets back to top, the window goes behind it, and a "leaveEvent" is
fired.

A .5 sec timeout seems to be a reasonable timeout also in different situations.


HenriWahl:
I'll check when back from vacations.

Issue #394, opened 2017-06-27
New plugin to support Monitos3 from Freicon


studost:
Freicon offers a ITSM and Monitoring solution based on Nagios. With this new plugin Nagstamon can display status information from Monitos3 instances. The plugin is based on the Nagstamon Livestatus plugin from MK.

https://itsm.freicon.de

PR #395, opened 2017-06-27
Nagstamon plugin for Monitos3

studost:
#394 New Nagstamon plugin to support Monitos3 from Freicon

Based on Nagstamon plugin Livestatus by MK.


HenriWahl:
I'll check when back from vacations.

Issue #409, opened 2017-07-27
[Feature request] Sensu Support


jothoma1:
Hi,
We are using sensu as our monitoring framework.
My team was using nagstatmon and really loves it :) so they ask me if nagstatmon can be used with sensu
Do you plan to add it ?
Thanks in advance

HenriWahl:
I do not plan it but if somebody has some code to add it to the current servers I would not mind to check it. :-)

vtsingaras:
@jothoma1 I have done an initial implementation. Can you test by cloning and running my fork at https://github.com/interworkscloud/Nagstamon/tree/sensu-support or alternatively using the installer https://github.com/interworkscloud/Nagstamon/releases/download/v0.0.1/Nagstamon-2.1-20170811-win64_setup.exe ? 

Note: The installer is clean but I take no responsibilities, if possible try to use the source directly.

HenriWahl:
This was fast... looks good!

HenriWahl:
I compiled all binaries, available now at https://nagstamon.ifw-dresden.de/files/incoming/

jothoma1:
@vtsingaras Thanks a lot !! i'm trying it but can you explain how to configure it ?
I dont really understand and have tried a lot of parameters but it dont work

vtsingaras:
@jothoma1 You can also set the base url of Uchiwa and if you have multiple uchiwa datacenters specify the one you want in the field monitor datacenter.

PR #419, opened 2017-08-14
Initial support for Sensu+Uchiwa

vtsingaras:
Fixes: https://github.com/HenriWahl/Nagstamon/issues/409


HenriWahl:
Can you check https://github.com/HenriWahl/Nagstamon/issues/436 to find out if it would work without the typing module?

PR #420, opened 2017-08-14
Fix parsing server name if the name starts with 'server_' or a left s…

vtsingaras:
…ubstring of it (eg. 'se')


HenriWahl:
Ah.. yes, of course. Thanks!

Issue #438, opened 2017-09-14
Nagstamon 3.0.0 crashes after start


bunnu:
Today I upgraded Nagstamon to version 3.0.0. Nagstamon shows a green icon after start and crashes immediately. Windows eventlog shows following information:
```
Name der fehlerhaften Anwendung: Nagstamon.exe, Version: 0.0.0.0, Zeitstempel: 0x00000000
Name des fehlerhaften Moduls: ucrtbase.dll, Version: 10.0.15063.413, Zeitstempel: 0x5ba8b66e
Ausnahmecode: 0xc0000409
Fehleroffset: 0x00000000000734be
ID des fehlerhaften Prozesses: 0x28e8
Startzeit der fehlerhaften Anwendung: 0x01d32d20bdbe8bf9
Pfad der fehlerhaften Anwendung: C:\Program Files\Nagstamon\Nagstamon.exe
Pfad des fehlerhaften Moduls: C:\WINDOWS\System32\ucrtbase.dll
Berichtskennung: e83cdefa-a0a2-42eb-b785-0b270d597a84
Vollständiger Name des fehlerhaften Pakets: 
Anwendungs-ID, die relativ zum fehlerhaften Paket ist: 
```

I'm using Nagstamon on Windows 10, x64 with Zabbix.

HenriWahl:
Do you use TLS encryption?

HenriWahl:
Are you able to run Nagstamon in a terminal following the info at https://nagstamon.ifw-dresden.de/docs/requirements/?

bunnu:
I don't use TLS encryption.
I'm able to run Nagstamon from source without application crash.

HenriWahl:
Can you enable debugging in Nagstamon hoping to see any output?

bunnu:
I just realized, the problem also occures while running Nagstamon from source.
As soon there is a issue with state 'warning', Nagstamon crashes on startup as well as during normal runtime. There is no problem with issues in state 'average'.

Here are the last lines of debug output:

```
Traceback (most recent call last):
  File "C:\Nagstamon\Nagstamon\QUI\__init__.py", line 3710, in get_status
    status = self.server.GetStatus()
  File "C:\Nagstamon\Nagstamon\Servers\Generic.py", line 821, in GetStatus
    status = self._get_status()
  File "C:\Nagstamon\Nagstamon\Servers\Zabbix.py", line 206, in _get_status
[...]

Issue #425, opened 2017-08-17
Error while configuring Nagstamon with zabbix 3.2


BernaldoPenasAntelo:
Hi:

I'm triying to configure Nagstamon with zabbix 3.2 and i'm getting:

Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/Nagstamon/QUI/__init__.py", line 3538, in get_status
    status = self.server.GetStatus()
  File "/usr/lib/python3/dist-packages/Nagstamon/Servers/Generic.py", line 801, in GetStatus
    status = self._get_status()
  File "/usr/lib/python3/dist-packages/Nagstamon/Servers/Zabbix.py", line 210, in _get_status
    trigger['application'] = this_item[0]['applications'][last_app]['name']
IndexError: list index out of range
Abortado

My config file is:

[server_Monitor server]
monitor_url = http://192.168.0.32/zabbix
use_display_name_host = False
enabled = True
use_autologin = False
timeout = 10
authentication = basic
password =
username = eJwFwd0KgyAYANAH6kJX29CLXciHstKvH0rodkgIRUizGPr0O+fFiEpRCRUkP/aPMw+B6UC4nhYbSabbfdA/4oFa48LmR51rHkswlxfQezg7WNvOVm2qoqu/OveFzDFIXbJhQnwzWgRO7Xzm1Cx/VaEiuQ==
monitor_cgi_url = http://192.168.0.32/zabbix
show_options = False
use_proxy = False
no_cookie_auth = False
autologin_key = eJyztS1P9LfI8HMMTQ82KEosLC139PUyMfD2T3IJ9gvXtzAK8SgqyC5PNnBzCan0y7L0Dw2oKg41T/EJMU4xDstzqUyJqKqq9EoFAFgKGBw=
proxy_username = eJwFwcEOgiAAANAP8qBLZOvQwXSaRJgIatyqIROmpEuZfX3vnQAvGPxd2nBX9jAG9FzU7+/nunm4GrZ8TVFK6pZmZM41tj6kD4oStgJPYOUvAQpMlAhYSjdPJTBsi3oOw25E8q7dcASS9ZaSXdWZaNxkb70rXkvDuqeOSczjtCoNd5lD8g/cJy1Q
proxy_address = http://proxyserver:port/
use_display_name_service = False
type = Zabbix
[...]

Issue #391, opened 2017-06-21
2.1-20170323 - Zabbix - Trigger description shown as service name


Metrax:
Hi,

not sure if this is a bug or a feature ;-)

Nagstamon 2.1-20170323 is showing the description of the trigger as service name when it is set.
For trigger without description it show the application name if the item. For me the application name makes the most sense here.

![grafik](https://user-images.githubusercontent.com/1061866/27378954-ccec1992-5679-11e7-8c49-7870416a0982.png)

![grafik](https://user-images.githubusercontent.com/1061866/27379037-14000794-567a-11e7-9ed3-aee92c3c85b5.png)

Regards,

Robert

mprenditore:
Hi,
It was meant to be a feature to have more detailed information about the alarm.
I would like to add a setting to let user to choose it but I'm a little bit busy ATM...
In case that annoy you, you can change the following code and recompile it.

**_Nagstamon/Servers/Zabbix.py:256_**
overwrite:
```
                    if len(service['comments']) == 0:
                        srvc = service['application']
                    else:
                        srvc = self.nagiosify_service(service['comments'])`
```
with:
```
#                 if len(service['comments']) == 0:
#                        srvc = service['application']
#                 else:
#                        srvc = self.nagiosify_service(service['comments'])
                    srvc = service['application']
```

then recompile it or reinstall it vie the setup.
I hope that will help.

mprenditore:
I'm not sure if it will be easy to implement... 
ATM I'm working on fixing the ACK on Zabbix, after I'll pass to add the setting for the description/application and after I'll look about the over popup :)

Thanks for the suggestions,

Stefano

HenriWahl:
Is this fixed with latest stable version from https://nagstamon.ifw-dresden.de/download/#stable?

mprenditore:
Not yet, I hope to finish it soon

mprenditore:
Now should be fixed! Keep me updated please

PR #440, opened 2017-09-15
Fixed crash in case of Zabbix Web Scenario

mprenditore:
Commit to solve the issues #438, #425 and probably even #391.


HenriWahl:
Thanks for the fast fix!

mprenditore:
Thanks to you for the fast merging and the new Major version :)
I'll need your help to understand how I can fix the ACK of Zabbix, probably I'll open a new issue late today.

Issue #374, opened 2017-04-07
SOLVED [Icinga] Quoting/escaping special characters in $status-info$ (email/url) - merged in #486


e1ze:
Hi,

is it possible to quote the $status-info$ for usage in url actions such like email? I have a problem if there is a special character like % in the status information of icinga, it cannot be displayed in an email with the url command.

Example -> ? should be a %:
Service: sys_cpu_remote_usage_average 
Status Information: **CRITICAL: 1: average load 99? warning, 5: average load 99? critical, 15: average load 99? critical** 
Another Example:
Status Information: **DISK WARNING - free space: /mnt/xxx 79298 MB (5?node=99?** 
This should be: **(5% inode=99%):**

Also if you have quotes (**""**) in your status information, the url command is not working and then you will get the error message that the command line argument is invalid.

Example -> quoted databse in pgsql WAL File Check from check_postgres:
Status information: **POSTGRES_WAL_FILES WARNING: DB "db" (host:10.x.x.x) WAL files found:** 

We have a lot of different checks and plugins and it's not possible for me to change every plugin or escape every output.

Thanks 

Greetings René

HenriWahl:
Sorry for the late answer but always much other stuff to do... what happens if you quote the $STAUTS_INFO$ variable in your action string?

e1ze:
Ah... Shit, I totally forgot this issue but better late than never I just tested it.
If i use quotes for "$STATUS-INFO$" i get an invalid response (wrong command line argument).
Thanks

e1ze:
Issue was solved: [Pull Request #486 ](https://github.com/HenriWahl/Nagstamon/pull/486)

PR #486, opened 2018-04-18
feature: quoting variables in custom actions

DeadHunter:
added feature to quote variables with urllib.quote for custom actions. To use this, add 2nd "$" at the beginning and end of the variable. Example: $$STATUS-INFO$$

#374

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -41,6 +41,8 @@ import base64
 import datetime
 import traceback
 
+from urllib.parse import quote
+
 from collections import OrderedDict
 from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks for your contribution... but does it now add the extra $ to every placeholder variable?

DeadHunter:
i just added the additional replace before the original replace

```
                mapping = {'$HOST$': info['host'],
                        '$SERVICE$': info['service'],
                        '$ADDRESS$': info['address'], [...]

                # mapping mapping
                for i in mapping:
                    # mapping with urllib.quote
                    string = string.replace("$"+i+"$", quote(mapping[i]))
                    # normal mapping
                    string = string.replace(i, mapping[i])
```
it works for every placeholder you defined in "mapping". If it finds the placeholder with double "$" it will be replaced with the quoted value, otherwise the original value will be used.

You can use both:

```
mailto:servicedesk@my.org?subject=Monitor alert: $HOST$ - $SERVICE$ - $$STATUS-INFO$$&body=Please help!.%0d%0aBest regards from Nagstamon
```

HenriWahl:
Ah... OK. Sounds cool!

PR #492, opened 2018-05-22
Enhanced zabbix backend supporting host name, acknowledge & maintenance

vsalvador:
This changes allow support for host visible name, better trigger
identification and support for host maintenance and trigger acknowledged
status, so triggers can be filtered.
Also, acknowsledge method works now in contextual menu
TODO: Host or event downtown setting from contextual menu


HenriWahl:
I added new binaries with your changes at https://nagstamon.ifw-dresden.de/download/

mprenditore:
Well done :)

vsalvador:
Thank you very much.

I've started to work on supporting downtime setting in zabbix.

When finished, I'll send you another pull request.

Thanks again.

Vicente

PR #502, opened 2018-08-02
Fixing crashing in case Zabbix service name have 3 characters

mprenditore:
fixing bug introduced on 038fa34

Due to the usage of Zabbix Service name in **attempts**, nagstamon
would have crashed  because of a string split needed  to check attempts


mprenditore:
@HenriWahl  Thanks :)

mprenditore:
Can you please compile it when you have time? I've started to work on a Mac lately but I'm not able to compile it. I think the problem is that my disk is partitioned with the "case-insensitive" flag.

HenriWahl:
See https://nagstamon.ifw-dresden.de/files/incoming for latest freshly compiled binaries.

mprenditore:
Thank you buddy! Cheers 👍

rafaelrs:
I've found problem with that fix. 

I have filter acknowledged problems in Nagstamon, so if i acknowledge problem it disappears from alerts.

But if such problems haves service name (application name in terms of Zabbix) with 3 characters, it's no longer filtered anymore and stays in alerts in Nagstamon and tray icon is non green.

I did debug of Nagstamon python code and found that 3 charater service name caused that behaviour and found that code the fix reason of that. 

I renamed my service name and wrong behaviour fixed.

PR #503, opened 2018-08-06
Fixing bug if item doesn't have values

mprenditore:
That can happen with webchecks


HenriWahl:
Latest binaries available at https://nagstamon.ifw-dresden.de/files/incoming/ again.

mprenditore:
Thank you buddy! now it works like a charm 👍

PR #533, opened 2019-01-08
change monoselection behavior to multiselection

Coolgeek789:
This a PR to allow multiple selection in a server :
It can handle : recheck, ack and downtime.
All other command are single selection only (as before).
I also added the ability to copy from the keyboard (ctrl+c) from all the selected line

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -252,7 +252,6 @@ APP.setStyleSheet('''QToolTip { margin: 3px;
 
 
 class HBoxLayout(QHBoxLayout):
-
     """
         Apparently necessary to get a HBox which is able to hide its children
     """
@@ -284,7 +283,6 @@ class HBoxLayout(QHBoxLayout):
 
 
 class SystemTrayIcon(QSystemTrayIcon):
-
     """
         Icon in system tray, works at least in Windows and OSX
         Several Linux desktop environments have different problems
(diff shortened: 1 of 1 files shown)


HenriWahl:
Looks cool... how exactly one can select multiple hosts/services? Pressing some key additionally?

Coolgeek789:
left clic for selecting one or more (shift to select a range, ctrl to select one by one)
right clic for the popup menu once the selection is made

HenriWahl:
OK... does not fully work here but looks promising :-)

Coolgeek789:
what doesn't work ?

HenriWahl:
right click for menu when closing window at hovering out:

`Traceback (most recent call last):
  File "/Nagstamon/Nagstamon/QUI/__init__.py", line 3330, in mouseReleaseEvent
    self.cell_clicked(index)
  File "/Nagstamon/Nagstamon/QUI/__init__.py", line 3479, in cell_clicked
    if self.miserable_service == 'Events':
AttributeError: 'TreeView' object has no attribute 'miserable_service'
`

Coolgeek789:
yeah. I don't have check_mk so it slipped. I will change that and make a new PR

HenriWahl:
No problem.

Coolgeek789:
Is it OK if I make this entry in the menu available only if one row is selected ?

HenriWahl:
Just do it - I will check later.

PR #548, opened 2019-02-20
monitos4x: host/services acknowledged fix

froehl:
monitos4x: host/services acknowledged fix for filters


froehl:
@HenriWahl Thx!

Issue #547, opened 2019-02-19
Latest build does not show custom actions for Nagios checks that are showing as down


chaoscreater:
If there's a host that is shown as down in Nagios, Nagstamon does not show any custom actions in the context menu. If you click on any other service or hosts, the custom actions will show up fine.

![image](https://user-images.githubusercontent.com/18227319/52988914-0f544400-3466-11e9-8bae-9c69ebc8fb68.png)

HenriWahl:
Yes, I can confirm this and it is... a bug. Will check this soon.

HenriWahl:
Please check latest available testing binaries.

PR #555, opened 2019-03-08
Fix for missing custom actions for hosts (#547)

Massl123:
I think i have found the issue with #547.
Please check if this is correct. I might have misunderstood the value here.
But with this change the actions are shown and the regular expressions for actions still seem to work.


HenriWahl:
@Massl123 can you check https://github.com/HenriWahl/Nagstamon/issues/558?

Massl123:
> 
> 
> @Massl123 can you check #558?

Yes, see #559

PR #561, opened 2019-03-17
Sensu and standalone checks causes Nagstamon to crash

gtufte:
new_sevice.site and status set to string so that the join function in Objects.py is satisfied and will not crash since it expects strings. POSTing to the Sensu API to recheck a standalone check is by design not supported and would cause Nagstamon to crash when trying to recheck all services


HenriWahl:
Thanks!

PR #562, opened 2019-03-18
Rechecking the Sensu keepalive service makes Nagstamon crash

gtufte:
Asking the SensuAPI to recheck the keepalive service makes Nagstamon crash, results for this service must come from the client. Also, you shouldn't have to be running Nagstamon in debug mode to be able to avoid crashing Nagstamon when testing/rechecking standalone services


HenriWahl:
👍

PR #567, opened 2019-04-15
Validation for the upcoming Centreon 19.04

BenoitPoulet:
Hello,

I run tests against the upcoming Centreon 19.04, everything seems fine.
So i wrote something in the code for debug.


HenriWahl:
Latest version 3.3-20190415 comes now with your changes.

PR #587, opened 2019-08-01
Fix missed mouse out to hide status window

ponchofiesta:
When moving the mouse too fast in and out of the indication window the status window is not hidden correctly and still shown. You have to move the mouse again over and out of the status window slowly.

This PR adds a timer that checks the mouse position and hides the window if it is not under the mouse.

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -1604,6 +1604,29 @@ class StatusWindow(QWidget):
                 if conf.icon_in_systray or conf.statusbar_floating:
                     self.hide_window()
 
+            # If the mouse cursor drives too fast over and out the window will not be hidden.
+            # Thus we check again with this timer to catch missed mouse-outs.
+            self.periodically_check_window_under_mouse_and_hide()
+
+    def periodically_check_window_under_mouse_and_hide(self):
+        """
+        Periodically check if window is under mouse and hide it if not
+        """
+        if not self.hide_window_if_not_under_mouse():
+            QTimer.singleShot(1000, self.periodically_check_window_under_mouse_and_hide)
+
+    def hide_window_if_not_under_mouse(self):
+        """
+        Hide window if it is under mouse
+        """
+        mouse_x = QCursor.pos().x()
+        mouse_y = QCursor.pos().y()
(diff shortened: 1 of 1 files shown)


ponchofiesta:
OK, I'm using QTimer instead of threading.Timer now. Do I need @pyqtSlot annotation? It works without so far.

HenriWahl:
Cool - Looks better this way I thinkm. @pyqtslot is unnecessary because your method is not used in a signal-slot-context.

HenriWahl:
Would be interesting if this will fix the similar problem with running Nagstamon in a Wayland session as XWayland application being displayed over a real Wayland application like Gnome Terminal. There it never closes because it got no notice that the pointer left the XWayland window.

ponchofiesta:
I have neither and can not test it but I think so.

HenriWahl:
Now available at https://nagstamon.ifw-dresden.de/download/#testing

HenriWahl:
I found something: can you please build in a check, if there is a context menu open? Otherwise the window closes while selecting something from action context menu... :-(

ponchofiesta:
Currently I have a workaround using `qApp.widgetAt(mouse_pos)`. But this only works as long as the mouse is over the context menu (or drop down list).

I did not find a way to check whether the window has any blocking windows open (like modal windows, context menus, drop down list). Thus every single trigger event must be catched to save the state of opened windows. Any other ideas?

HenriWahl:
I think I remember to use some lock to avoid refreshing of the view while a menu is open - might be recycled there. If I find the time to look for it I might do - maybe you are faster :-)

HenriWahl:
I found another bug: my: my default setting is "close if clicking somewhere", but now the popup closes every time the mouse is moved out. I think the setting should be respected.

ponchofiesta:
Yes, I noticed it too. :) I already fixed it in my local development branch. And I think I've found a fix for the other problem...

HenriWahl:
:+1:

ponchofiesta:
see #596

ponchofiesta:
Yes, I'll check that.

HenriWahl:
I fixed it with https://github.com/HenriWahl/Nagstamon/commit/87e6ef08fe11833eeb8f685c0026c13a73b0cf94

PR #596, opened 2019-09-20
Fix missed mouse out to hide status window (2)

ponchofiesta:
This adds the conf check for hiding window on mouse leave and checks if a context menu or drop down list is opened.

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -1606,8 +1606,10 @@ class StatusWindow(QWidget):
 
             # If the mouse cursor drives too fast over and out the window will not be hidden.
             # Thus we check again with this timer to catch missed mouse-outs.
-            if not conf.fullscreen and not conf.windowed:
-              self.periodically_check_window_under_mouse_and_hide()
+            if conf.close_details_hover \
+                    and not conf.fullscreen \
+                    and not conf.windowed:
+                self.periodically_check_window_under_mouse_and_hide()
 
     def periodically_check_window_under_mouse_and_hide(self):
         """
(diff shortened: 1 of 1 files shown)


Nagstamon/QUI/__init__.py
```diff
@@ -1620,9 +1622,11 @@ def hide_window_if_not_under_mouse(self):
...
         Hide window if it is under mouse
         """
-        mouse_x = QCursor.pos().x()
-        mouse_y = QCursor.pos().y()
-        if statuswindow.geometry().contains(mouse_x, mouse_y):
+        mouse_pos = QCursor.pos()
+
+        # Check mouse cursor over window and an opened context menu or dropdown list
+        if statuswindow.geometry().contains(mouse_pos.x(), mouse_pos.y()) \
+                or not qApp.activePopupWidget() is None:
```
HenriWahl: Where does qApp come from? Shouldn't it be APP as defined in line 30?

ponchofiesta: It is `QtWidget.qApp`. You have already imported everything from `QtWidget`. I assume `APP` is the same here.
https://doc.qt.io/qt-5/qapplication.html#qApp

HenriWahl: Yes, APP is the current instance there.

HenriWahl:
My shell's history lured me to do a forced git **push** instead a **pull** and so I overwrote you merged PR. Can you please request again?

ponchofiesta:
You already remerged it. You could have rebased on my feature branch to get the correct commits again. Or cherry pick the missing commits.

HenriWahl:
Yes, I included your changes as a patch file because they were not visible for me anymore after my forced push - either way, now they are in master.

HenriWahl:
There is another bug or at least strange looking behaviour. The status window does not open or closes too fast when hovering the pointer in the floating statusbar in a Fedora Wayland session. Can you confirm this? Session is a default GNOME Wayland session.

ponchofiesta:
Sorry, can't verify. I don't have any of this software installed.

HenriWahl:
I added a simple Wayland detection and disabled your patch in case of running in Wayland. Maybe there might be a more sophisticated solution but as there are problems with Wayland in general I think it doesn't matter.

HenriWahl:
Me again.

Could it be that the problem you try to solve is caused by https://github.com/HenriWahl/Nagstamon/blob/master/Nagstamon/QUI/__init__.py#L1665?

Apparently your fix causes new trouble in Windows, making the floating statusbar jump along the screen from time to time...

PR #597, opened 2019-10-10
Fix missed mouse out to hide status window when systray icon is used

ponchofiesta:


ponchofiesta:
You already fixed it :)

PR #599, opened 2019-10-15
monitos 4.4+ no longer experimental: A lot tested with version 3.3-20…

froehl:
monitos4 no longer experimental: A lot tested with version 3.3-20191010 (Win10 & macOS 10.14)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 Nagstamon
 =========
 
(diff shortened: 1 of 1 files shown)


HenriWahl:
OK... what is by the way the difference between monitos 3 and 4? I monitos 3 obsolete or has it still its userbase?

froehl:
> OK... what is by the way the difference between monitos 3 and 4? I monitos 3 obsolete or has it still its userbase?

The differences between monitos 3 and monitos 4 are a lot ;-). New Major version. Among others monitos 4 has new APIs. There is still a userbase for monitos 3, but shrinking.

Issue #154, opened 2015-08-19
Add "Use Expire Time" for an acknowledgement


neufeind:
In Icinga an acknowledgement has the option "Use Expire Time". With that you can accept an alarm for a given time.

If it returns to okay-state before, then the monitoring takes effect immediately again (in contrast to if you specify a downtime of some hours). But if the alarm doesn't go away after the given time the alarm will return.
Example usecase: When sending out a large newsletter you can ack that for a certain time a high load on the mailqueue is okay.

HenriWahl:
Sounds interesting.

idl0r:
I'd like to see that option too. Would be really handy.

Mahagon:
This would be really usefull

Dreamfire78:
+1 for this feature from me. Now I have to open the WebInterface everytime I have to add an expiration time.

HenriWahl:
Just 4 years later this feature arrived - thanks @Massl123 

Check at https://nagstamon.ifw-dresden.de/download/#testing

PR #602, opened 2019-11-01
Added use expire time feature for acknowledgements to IcingaWeb2

Massl123:
This adds #154

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -6164,6 +6164,16 @@ class Dialog_Acknowledge(Dialog):
     def __init__(self, dialog):
         Dialog.__init__(self, dialog)
 
+        self.TOGGLE_DEPS = {
+            self.ui.input_checkbox_use_expire_time: [self.ui.input_datetime_expire_time]
+            }
+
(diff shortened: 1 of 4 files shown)


HenriWahl:
Sorry for the late reply... I am about to get a stable release 3.4 ready and there are still some other small things to fix. I will look into your changes after this.

Massl123:
Will this be included in 3.4? I think it would be a nice addition for that release.

neufeind:
What does it take to get this working with an Icinga 1.x-installation as well? There is also an expiry-time when setting an ack. Inputfields on the webinterface in the HTML have these names:

<input type="checkbox" id="expire_checkbox" name="use_ack_end_time" onclick="if (document.getElementById('expire_checkbox').checked == true) document.getElementById('expired_date_row').style.display = ''; else document.getElementById('expired_date_row').style.display = 'none';">

<input type="text" class="timepicker hasDatepicker" name="end_time" value="12-02-2019 20:30:20" size="25" id="dp1575307821023">

Massl123:
Should be pretty easy. If i find the time i can make an PR.
But this will be included in the release after 3.4.

Issue #609, opened 2019-11-20
Passwords in debug log


mbirth:
Version: 3.3-20191114

I've noticed the debug log contains server passwords in cleartext. They should be replaced by something like `*****` or redacted otherwise so no actual password ends up in the logfile.

PR #610, opened 2019-11-25
Allow string values to be looged and fix FetchURL function

Massl123:
File "/usr/lib/python3/dist-packages/Nagstamon/Servers/Generic.py", line 1421, in FetchURL
    for key in cgi_data_log.keys():
AttributeError: 'str' object has no attribute 'keys'

Some servers send cgi_data as strings - e.g. Nagios when doing the recheck action.
Related to #609


HenriWahl:
Thanks!

PR #611, opened 2019-11-26
Add expire time for acknowlegements to settings and other fixes

Massl123:
- fix setting expire time when selecting "acknowledge all services on host"
- add expire time to settings for actions and to configuration file
- fix resizing of acknowledgement dialog when changing between server with expire time and one without

Btw, nice way of managing configuration and settings dialog, really easy to expand!


HenriWahl:
Do you plan more PRs? :-)

If not I think we are going to reach stable release 3.4 soon.

HenriWahl:
I added a dection if an IcingaWeb2 exists at all - if not, your new default option is not shown to avoid confusion.

Massl123:
For now no more PRs are planned, unless i find a bug.

> I added a dection if an IcingaWeb2 exists at all - if not, your new default option is not shown to avoid confusion.

Great idea!

PR #621, opened 2020-01-09
[Centreon] Fix ack when using autologin key

BenoitPoulet:
The centreon token was not gathered so the ack was not effective.


HenriWahl:
Available at https://nagstamon.ifw-dresden.de/download/#testing

PR #630, opened 2020-02-29
Momiji patch 1

momiji:
Hello,

This PR to change the zabbix login logic.

It appears that when zabbix server is not yet available, which happens if program starts before network  is connected (when I'm working with VPN for example), then login fails and is never called later, unless restarting program or doing a fake server update.

This patch allows to retry login until it succeeds, and fixes the "unable to decode" error to return an exception instead of exiting ; this happens during the very start of the VPN connection, due to stupid enterprise hidden reverse proxy...

Do not hesitate if you need more information.

Regards.


HenriWahl:
Thanks!

PR #636, opened 2020-04-21
Adding support for SOCKS5 Proxy

justino:
When using a SOCKS5 proxy, simply set the URL to `<socks scheme>://<host>:<port>`.
Adds the following SOCKS schemes: `socks5`, `socks5h`
This requires SOCKS support in the `Requests` module using the `pysocks` module.
Added new dependencies for Debian and Redhat
Tested in Debian

I don't have access to a Redhat box to test the build process.
I also have no idea how to build this for Windows and add the new required module.


HenriWahl:
That's it? Already works?

justino:
The only thing that needs updating is the docs found at https://nagstamon.ifw-dresden.de/docs/requirements/

`pysocks` is now a requirement. So it now needs to be listed in all the pip commands, and listed in top list of pip packages

HenriWahl:
Is this extra package available for Debian and Fedora too?

justino:
Yes, it is.
In Debian: `python3-socks`
In Fedora: `python3-pysocks`

In the PR, I updated the appropriate build systems accordingly.

HenriWahl:
I updated all relevant parts.

The latest binary testing version contains your addition too.

PR #645, opened 2020-06-10
Prometheus: support non-kubernetes alerts and a bit of cleanup

foosinn:
This adds support for alerts that have neither `pod_name` nor `namespace` defined.

Also removes the hardcoded `pod_name` in the browser url, since this breaks all alerts that aren't pod related.

Also some changes to make the linter happy and the code easier to read.


HenriWahl:
Thanks!👍

PR #647, opened 2020-06-12
Added configuration for Prometheus mappings

stearz:
Finally I managed to add the front-end modification so that users now can configure their own Prometheus labels/annotations to be mapped as Nagstamon fields: hostname, servicename / status_information.
I preset the values in config.py with a (for me) sane default.


HenriWahl:
Thanks, will be built asap.

HenriWahl:
Available at https://nagstamon.ifw-dresden.de/files/incoming/

PR #657, opened 2020-08-05
Toggle popup on system tray middle click

shastry:


HenriWahl:
Funny idea, but why not... thanks!

Issue #655, opened 2020-08-03
VERSION_ID unavailable on Gentoo in Helpers.py


arabek:
In line 457 of Nagstamon/Helpers.py:
```python
                    os_release_dict.get('VERSION_ID').lower(),
```

The VERSION_ID verb is unavailable in the os-release file on Gentoo:
```
# cat /etc/os-release
NAME=Gentoo
ID=gentoo
PRETTY_NAME="Gentoo/Linux"
ANSI_COLOR="1;32"
HOME_URL="https://www.gentoo.org/"
SUPPORT_URL="https://www.gentoo.org/support/"
BUG_REPORT_URL="https://bugs.gentoo.org/"
```
making it impossible to compile the package on Gentoo with Python >=3.7.

PR #660, opened 2020-08-14
Fix issue #655 - by returning 'unknown' as VERSION_ID if value missing

simoncadman:
Just added code to set VERSION_ID to a default value if it doesn't exist, fixes using python 3.7+ on Gentoo where they don't set a VERSION_ID in /etc/os-release (related to https://bugs.gentoo.org/718424). Possibly affects more than Gentoo as https://www.freedesktop.org/software/systemd/man/os-release.html says VERSION_ID is optional.


HenriWahl:
Good idea - you could even save some lines when setting the default value in the .get() method of the dictionary.

simoncadman:
Commit updated to return default value on .get , thanks.

Issue #618, opened 2020-01-04
scaled display results in very small fonts (Win10)


hknet:
Hi
on Win10 with highres-displays it is quite common to have fonts and other elements scaled to 150 or even 200% to keep things readable.

nagstamon 3.4 does obviously not take this scaling into account and makes itself quite hard to read:
![image](https://user-images.githubusercontent.com/4014716/71759264-c4d78c80-2eaa-11ea-89c6-fedae5685c78.png)

HenriWahl:
As a partly mitigation you can change don't size.

In theory the underlaying Qt should handle high res displays.

hknet:
well, tried that :)
changed fontsize from 6 to 9 -> resulted after "ok" in a killed nagstamon process :)
restarted nagstamon and changed the fontsize again - this time it did not kill itself:

![image](https://user-images.githubusercontent.com/4014716/71764936-aa270700-2eee-11ea-91eb-804d2d8d69bf.png)

yet the result does change the fontsize only for the reported hosts and services, not for any other fonts, also this is only a very cumbersome workaround as usually multi-display-environments have different scalings on different displays and therefore this is not a real solution.
I don't know about the Qt but it obviously does not scale correctly :)

HenriWahl:
I found some information about Qt + HiDPI and will try to implement this as soon as possible.

HenriWahl:
Please check 3.5-20200105 from https://nagstamon.ifw-dresden.de/download/#testing

hknet:
Hi, thank you for your quick response, I now tried 20200108 and the results are better, yet scaling probably will need to be done for any text-element:

![image](https://user-images.githubusercontent.com/4014716/72104489-d34ef980-332b-11ea-9c63-38f9b0182315.png)


on the other hand the funny effect when moving nagstamon from a HiDPI display to a "normal" DPI display it gets a bit confused there:
![image](https://user-images.githubusercontent.com/4014716/72104605-1610d180-332c-11ea-9340-9172b14e3cfc.png)
if I do a refresh after this display appears the windows gets a normal positioning as well as correct dimensions to see something :)

HenriWahl:
Well, I guess this is Qt's part... I could not do more than activate HiDpi-awareness at all. :-(

PR #661, opened 2020-08-17
fix AttributeError for Qt.AA_EnableHighDpiScaling with Qt 5.5

oliver:
The Qt.AA_EnableHighDpiScaling property was introduced with Qt 5.6.

Without the fix I'm getting this error message when starting Nagstamon on a Linux system with Qt 5.5:

```
  File "/usr/bin/nagstamon", line 45, in <module>
    from Nagstamon.QUI import (APP,
  File "/usr/lib/python3/dist-packages/Nagstamon/QUI/__init__.py", line 117, in <module>
    QApplication.setAttribute(Qt.AA_EnableHighDpiScaling, True)
AttributeError: type object 'Qt' has no attribute 'AA_EnableHighDpiScaling'
```

The line which causes the exception was added in https://github.com/HenriWahl/Nagstamon/issues/618. I suppose it's no big problem if HighDPI doesn't work on such old systems, though.

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -114,7 +114,10 @@ if not OS in OS_NON_LINUX:
         DBUS_AVAILABLE = False
 
 # enable HighDPI-awareness to avoid https://github.com/HenriWahl/Nagstamon/issues/618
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks!

PR #670, opened 2020-10-08
Allow to install on nixos

foosinn:
This fixes `get_distro()` and the path resolution for `RESOURCES` on nixos and maybe other platforms.

Since these are just fallback options this should not affect any currently working systems.

### DIFF
--- a/Nagstamon/Config.py
+++ b/Nagstamon/Config.py
@@ -1064,3 +1064,6 @@ except Exception as err:
         if os.path.exists(path):
             RESOURCES = path
             break
+    else:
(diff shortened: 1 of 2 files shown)


Nagstamon/Config.py
```diff
@@ -1064,3 +1064,6 @@ def __init__(self, **kwds):
         if os.path.exists(path):
             RESOURCES = path
             break
+    else:
+        from pathlib import Path
```
HenriWahl, line 1068: Shouldn't this be imported at the head of the file?

PR #671, opened 2020-10-20
add filter for unreachable services

jabdoa2:
- Add `(UNREACHABLE SERVICE)` to the status line of unreachable services
- Add a new filter checkbox to filter unreachable services
- Filter out unreachable services when checkbox is checked
- Query `is_reachable` field for icinga2web

### DIFF
--- a/Nagstamon/Objects.py
+++ b/Nagstamon/Objects.py
@@ -147,6 +147,7 @@ class GenericService(GenericObject):
 
     def __init__(self):
         GenericObject.__init__(self)
+        self.unreachable = False
 
     def get_host_name(self):
         return str(self.host)
--- a/Nagstamon/QUI/settings_main.py
+++ b/Nagstamon/QUI/settings_main.py
@@ -2,7 +2,7 @@
 
 # Form implementation generated from reading ui file 'settings_main.ui'
 #
-# Created by: PyQt5 UI code generator 5.13.1
+# Created by: PyQt5 UI code generator 5.13.2
 #
 # WARNING! All changes made in this file will be lost!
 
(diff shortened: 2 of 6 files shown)


jabdoa2:
@HenriWahl any chance to get this merged?

HenriWahl:
Sorry for the late reply - I am about to check this.

Nagstamon/Objects.py
```diff
@@ -147,6 +147,7 @@ class GenericService(GenericObject):
 
     def __init__(self):
         GenericObject.__init__(self)
+        self.unreachable = False
```
HenriWahl, line 150: Sorry for the late reply.

Why has self.unreachable to be initialized here? Shouldn't it be set like other properties?

Nagstamon/QUI/settings_main.py
```diff
@@ -1642,3 +1646,13 @@ def retranslateUi(self, settings_main):
         self.label_acknowledge_submit_result.setTitle(_translate("settings_main", "Submit check result:"))
         self.label_defaults_submit_check_result_comment.setText(_translate("settings_main", "Comment:"))
         self.tabs.setTabText(self.tabs.indexOf(self.tab_defaults), _translate("settings_main", "Defaults"))
+
+
+if __name__ == "__main__":
```
HenriWahl, line 1651: Where does this condition come from? Is it inserted by the GUI compiler pyuic5?

HenriWahl:
Stupid question: what are unreachable services? As far as I understand "unreachable" applies only to hosts?

Nagstamon/QUI/settings_main.py
jabdoa2: Exactly

Nagstamon/Objects.py
jabdoa2: Because it applies to services only. We can move it up if you want

jabdoa2:
> Stupid question: what are unreachable services? As far as I understand "unreachable" applies only to hosts?

Those are services with a dependency. For instance if a service is actually checked via Prometheus and that host is down it would fail. For those services we define a dependency so we can filter them out as unreachable.

PR #672, opened 2020-10-20
bail out if cacert cannot be found on Mac to prevent runtime errors

jabdoa2:
Currently, Nagstamon will fail silently when run from the source folder. This will at least make it explicit. Can you add the following to the install or dev install instructions on your website?

```
mkdir certifi
cd certifi
wget https://curl.haxx.se/ca/cacert.pem
```

Afterwards, Nagstamon works fine on a development system.

### DIFF
--- a/Nagstamon/Servers/Generic.py
+++ b/Nagstamon/Servers/Generic.py
@@ -200,6 +200,8 @@ class GenericServer(object):
                 # ...and its content
                 with open(self.cacert_path, mode='rb') as file:
                     self.cacert_content = file.read()
+            else:
+                raise AssertionError("Cannot find cacert.pem. SSL Connections will not work.")
 
         # Special FX
         # Centreon


HenriWahl:
Sorry for the late reply - can you explain what exactly is the purpose of this PR?

jabdoa2:
Currently nagstamon will silenty fail to connect to any SSL host if you omit the (undocumented) steps above when running it from the source folder. This at least issues a clear error.

HenriWahl:
OK, I remember the whole pyinstaller macOS thingy being a pretty insane but working approach - for production. You are right, might look in a devel setup kind of none-fitting.
I wunder anyway why the cacert.pem does not come with certifi in your case?

HenriWahl:
I don't know what I do wrong but it just works, running on macOS from source via HTTPS.
I set up a fresh git checkout and installed all requirements.
Are you sure there might not be another cause?

HenriWahl:
Looks like there is no need for this PR.

PR #678, opened 2020-11-06
Added Icinga2 API. Partially Closes #647

ser:
Added Icinga2 API monitoring which is the most important part for most of users. 

Additional features like submitting check results, acknowledgements, rechecks will be added in future versions as it requires a lot of time for testing of the code.

I have added two additional libraries as requirements, one is essential as it's an interface to API, second one is extremely helpful to limit datetime library inefficiency. 

Bug reports appreciated.

### DIFF
--- a/Nagstamon/Servers/__init__.py
+++ b/Nagstamon/Servers/__init__.py
@@ -25,12 +25,18 @@ import urllib.parse
 
 from collections import OrderedDict
 
+# we need this for uncommon modules
+try:
+    import icinga2api
+    icinga2api_is_available = True
+except:
+    pass
+
 # load all existing server types
 from Nagstamon.Servers.Nagios import NagiosServer
 from Nagstamon.Servers.Centreon import CentreonServer
 from Nagstamon.Servers.Icinga import IcingaServer
 from Nagstamon.Servers.IcingaWeb2 import IcingaWeb2Server
-from Nagstamon.Servers.Icinga2API import Icinga2APIServer
 from Nagstamon.Servers.Multisite import MultisiteServer
 from Nagstamon.Servers.op5Monitor import Op5MonitorServer
 from Nagstamon.Servers.Opsview import OpsviewServer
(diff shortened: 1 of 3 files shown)


HenriWahl:
Thanks for the contribution.

Can you include a check that if the Python module icinga2api is not available the Icinga2API won't too? I think of Linux distros which do not include the requirements by pip but via their respective package managements. Or can you find out if something like an icinga2api package is available for the mostl used distros?

ser:
Unfortunately icinga2api is unavailable in most of distros if any as a package :-(

HenriWahl:
> Unfortunately icinga2api is unavailable in most of distros if any as a package :-(

OK, then a simple is-icinga2api-module-available condition might be a good idea to avoid confusion for desktop users.

> Maybe it is time to prepare Nagstamon as an AppImage?

Rather I think adding it to the already existing flatpak at https://github.com/flathub/de.ifw_dresden.nagstamon because meanwhile flatpaks do work on almost every distribution too.

ser:
> > Unfortunately icinga2api is unavailable in most of distros if any as a package :-(
> 
> OK, then a simple is-icinga2api-module-available condition might be a good idea to avoid confusion for desktop users.

How do you propose to check it, for example ```pkgutil.find_loader("icinga2api")```? pkgutil module is available by default I suppose.

> > Maybe it is time to prepare Nagstamon as an AppImage?
> 
> Rather I think adding it to the already existing flatpak at https://github.com/flathub/de.ifw_dresden.nagstamon because meanwhile flatpaks do work on almost every distribution too.

It sound good either :+1:

HenriWahl:
This should to the trick: https://github.com/HenriWahl/Nagstamon/blob/ff9a31d002ce9e4904923100c760e6223ff3d074/Nagstamon/QUI/__init__.py#L6988 , no extra module needed.

ser:
I'm afraid not, sys module does know about the module only if it is loaded already. It would require to use try/except to know if it is available, which is far from perfect.

HenriWahl:
True, the example I sent you does not fully match.

But isn't a try/except combo just needed once at the beginning and later it could be decided via `if 'icinga2api' in sys.modules`?

ser:
OK, let me think and try :) I will report soon.

ser:
@HenriWahl what do you think about this? I had to swap tuple for a list as tuples are unchangeable.

Nagstamon/Servers/__init__.py
```diff
@@ -25,12 +25,18 @@
 
 from collections import OrderedDict
 
+# we need this for uncommon modules
+try:
+    import icinga2api
+    icinga2api_is_available = True
+except:
+    pass
```
HenriWahl, line 33: Here would be a good place to set icinga2api_is_available = False, otherwise it will crash later, right?

PR #692, opened 2021-02-03
Thruk: fix login against latest release

sni:
Cleaning up login handling. It is not required to detect cookie auth, it should simply work both ways.
Changes tested against latest Thruk and back till v2.10 which is already more than 5 years old.
Tested with and without cookie auth.


HenriWahl:
Thanks!

PR #694, opened 2021-02-07
fixing acknowledged services filter in zabbix

### LINKED ISSUE #654: 3 character application name brakes acknowledger filter
I've found problem with that fix. 

I have filter acknowledged problems in Nagstamon, so if i acknowledge problem it disappears from alerts.

But if such problems haves service name (application name in terms of Zabbix) with 3 characters, it's no longer filtered anymore and stays in alerts in Nagstamon and tray icon is non green.

I did debug of Nagstamon python code and found that 3 charater service name caused that behaviour and found that code the fix reason of that. 

I renamed my service name and wrong behaviour fixed.

_Originally posted by @rafaelrs in https://github.com/HenriWahl/Nagstamon/pull/502#issuecomment-656377374_

dizixagi:
Fixes #654

Skipping block of code that sets
`service.visible = True`
on all services with attempt field (populated by application name from zabbix item), when serwer type meches Zabbix.

### DIFF
--- a/Nagstamon/Servers/Generic.py
+++ b/Nagstamon/Servers/Generic.py
@@ -1088,7 +1088,8 @@ class GenericServer(object):
                             self.Debug(server=self.get_name(),
                                        debug='Filter: SOFT STATE ' + str(host.name) + ';' + str(service.name))
                         service.visible = False
-                else:
+                # fix for https://github.com/HenriWahl/Nagstamon/issues/654
+                elif self.TYPE.startswith("Zabbix"):
                     if len(service.attempt) < 3:
                         service.visible = True
                     elif len(service.attempt) == 3:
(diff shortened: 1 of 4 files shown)


Nagstamon/Servers/Generic.py
```diff
@@ -1088,7 +1088,8 @@ def GetStatus(self, output=None):
                             self.Debug(server=self.get_name(),
                                        debug='Filter: SOFT STATE ' + str(host.name) + ';' + str(service.name))
                         service.visible = False
-                else:
+                # fix for https://github.com/HenriWahl/Nagstamon/issues/654
+                elif not self.TYPE.startswith("Zabbix"):
```
HenriWahl, line 1092: Can you find a solution without changing Generic.py to keep it cleaner between the several server implementations?

dizixagi: Filtering in Generic.py (checking against configured filters and acknowledged/scheduled_downtime then setting visible to False) is done above this check against length of attempt. There literary not much more that can be done inside Zabbix.py that would be much more cleaner.
I can think of two different solutions:

1. Implement another filtering inside Zabbix.py - but that just redundant 
2. Pad every attempt with white spaces to be more then 3 characters in length - ugly hack.

HenriWahl: Well, I see... and anyway there are a bunch of server-specific conditions already in there. So it may be.

Hope one day find the time to clean all this old code up.

PR #706, opened 2021-03-30
Fix: Initialize `message` to prevent check uninitialized error

wolneykien:
A small startup fix to prevent this:

```
  File "/usr/lib/python3/site-packages/Nagstamon/QUI/__init__.py", line 6982, in check
    if message != '':
UnboundLocalError: local variable 'message' referenced before assignment
```


HenriWahl:
Thanks!

Issue #710, opened 2021-04-07
Custom action regular expression [3.6-rc2]


poldyaev:
Tell me what regular expression should I use to remove the port from host when using alertmanager?
I've tried using a `(\ S +):` expression, but I still get host: port in the output.
![DeepinScreenshot_nagstamon_20210407154951](https://user-images.githubusercontent.com/878061/113861988-e82d6d80-97b8-11eb-8957-b6f1a737e603.png)

matgn:
Same problem here! I tried regex groups but I think Nagstamon didn't implement it

PR #718, opened 2021-04-19
[Alertmanager server] omit port number for hostname

matgn:
With the Alertmanager Server, we can't use actions : Prometheus defines instance with port number. 
We can add this to omit port number and then use actions on the target (like ssh etc.)

The problem here :
#710
(Trying to workaround with regex)


HenriWahl:
Thanks - it is in the latest binaries at https://github.com/HenriWahl/Nagstamon/releases/tag/latest now.

PR #729, opened 2021-05-14
Feature/ecp authentication

guillomovitch:
Those two patches add support for ECP authentication, ie SAML for non-browser HTTP clients: https://wiki.shibboleth.net/confluence/display/CONCEPT/ECP

I tested it successfully with our own IdP, but I didn't invested much effort into error management.


HenriWahl:
Thanks!

PR #735, opened 2021-05-21
add requests-ecp dependency

guillomovitch:
I forgot this dependency in my previous PR.

As it will only concern very few people, a better solution would probably to make this dependency optional, and just avoid proposing ECP authentication if loading fail. The same could apply to Kerberos authentication, actually. However, if your main target audience is people using pre-built distributions with all dependencies included, or people building from sources fetching those dependencies directly from PyPi, and not people trying to minimize dependencies to required one only, this seems just useless technical complexity.

### DIFF
--- a/build/requirements/linux.txt
+++ b/build/requirements/linux.txt
@@ -7,4 +7,5 @@ pysocks
 python-dateutil
 requests
 requests-kerberos
+requests-ecp
 setuptools==44.1.1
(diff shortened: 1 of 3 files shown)


HenriWahl:
Thanks - lets see what happens. :-)
Did you spot the dependency as a package in Linux distros like Debian or Fedora?

PR #742, opened 2021-06-10
[Zabbix 5.4] Application name temporary workaround

mprenditore:
With Zabbix 5.4, application names have been transformed into `Tags` that means the object of the trigger doesn't have the app name anymore.

I didn't had enough time yet to investigate how to retrieve the Application tags via the API, but as a quick workaround, we can show `NO APP`.

This change is not going to cause issues to users with older versions of Zabbix.


HenriWahl:
Thanks!

PR #751, opened 2021-07-27
Fix SnagView interpreting all up and ok as pending

fabm3n:
If i execute this in python: print(int(0 or 4))
It always prints the number 4. So all Hosts and Services which are Up or Ok will be skipped because Nagstamon thinks they are pending (Status 4).


HenriWahl:
:+1:

Issue #755, opened 2021-08-04
3.7-20210518 Zabbix 5.4 Crash on schedule downtime


hunsbea:
![image](https://user-images.githubusercontent.com/5019061/128131891-9555cdaa-df28-4096-b97c-ce1ff41cafdb.png)

Nagstamon version 3.7-20210518
Zabbix version 5.4 latest

When I right-click a row in Nagstamon, select Schedule Downtime, and click OK, Nagstamon crashes, and does not create the downtime.

The funny thing is, it actually works ONCE. I mean, If there are currently NO scheduled down times in Zabbix, then this feature works and creates the downtime. But if there are any existing downtimes in Zabbix maintenance, then this fails, and Nagstamon crashes.

I enabled debug and debug logs, but Nagstamon does not write anything to the debug file, so I can't be any more help there.

This is a great great feature, so I hope it can be fixed soon! Let me know if you need any additional details.

HenriWahl:
Are yo able to run Nagstamon from source in console so we could get the actual error?

hunsbea:
I will try to set that up and will get back to you.

hunsbea:
Are there any instructions on how to build from source besides https://nagstamon.de/documentation ?

I've tried building on Windows 10 with Python 3.6, and I've tried building on Ubuntu 20.04 with Python 3.8, and both builds failed similarly.

On Windows, I did:
```
pip install -r build\requirements\windows.txt
python setup.py install
```
and I get:
```
C:\...\lib\distutils\dist.py:261: UserWarning: Unknown distribution option: 'executables'
  warnings.warn(msg)
C:\...\venv\lib\site-packages\setuptools\dist.py:485: UserWarning: The version specified ('3.7.20210809.') is an invalid version, this may not work as expected with newer versions of setuptools, pip, and PyPI. Please see PEP 440 for more details.
  "details." % self.metadata.version
running install
running bdist_egg
running egg_info
writing Nagstamon.egg-info\PKG-INFO
writing dependency_links to Nagstamon.egg-info\dependency_links.txt
writing top-level names to Nagstamon.egg-info\top_level.txt
reading manifest file 'Nagstamon.egg-info\SOURCES.txt'
writing manifest file 'Nagstamon.egg-info\SOURCES.txt'
installing library code to build\bdist.win-amd64\egg
running install_lib
running build_py
installing package data to build\bdist.win-amd64\egg
running install_data
[...]

PR #756, opened 2021-08-13
fix dup downtime crash + add app

hunsbea:
Addresses https://github.com/HenriWahl/Nagstamon/issues/755 and also goes a bit further.

So firstly, in Zabbix 5.4 if you send a maintenance request with a duplicate name, it will get rejected and crash Nagstamon. I fixed that with a try-except.

Is silently ignoring this failure the right thing to do? In my case, it is. If I've already created a downtime for this host+application, I don't want to create another one. The view of the host+application will be updated on the next refresh to show the downtime correctly.

This brings me to the point about applications. Currently, if you create a downtime for a host, it will automatically downtime all services on that host. I don't want that. In Zabbix 5.4, each Template has an Application tag for the items. This tag gets sent to the maintenance API to further control which triggers get silenced.

image

Finally, I fixed the "NO_APP" showing up in UI to use this Application tag value in the "attempt" column instead. It looks great.

image

I will likely make a build from this and use it internally, so this pull request is mostly a courtesy, and I hope it can be helpful. Feel free to edit it as you see fit.

Cheers.

### DIFF
--- a/Nagstamon/Servers/Zabbix.py
+++ b/Nagstamon/Servers/Zabbix.py
@@ -97,7 +97,14 @@ class ZabbixServer(GenericServer):
     def getLastApp(self, this_item):
         if len(this_item) > 0:
             if "applications" not in this_item[0]:
-                return "NO APP"
+                if 'tags' in this_item[0]:
+                    app = "NO APP"
+                    for tag in this_item[0]['tags']:
+                        if tag['tag'] == 'Application':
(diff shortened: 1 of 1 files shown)


hunsbea:
I just realized the "acknowledge" feature is also slightly broken, I'll add another commit here tomorrow.

HenriWahl:
Thanks!

Issue #743, opened 2021-06-13
macOS arm64/Apple Silicon Build


dotsam:
It looks like pyinstaller is adding support for both arm64/Apple Silicon builds for macOS, and also universal builds (although there are more caveats to get these universal2 builds working: https://pyinstaller.readthedocs.io/en/latest/feature-notes.html#macos-multi-arch-support). I just wanted to document what I did to get an arm64 build working locally in case others are looking to do the same.

To begin, I've installed python 3 and pyqt 5 from Homebrew (`brew install python@3.9 pyqt@5`).

Then I've installed pyinstaller from source, including re-compiling boot loaders, which creates x86/arm64 fat binaries: https://github.com/pyinstaller/pyinstaller/issues/5886#issuecomment-854116364

Finally, I cloned the Nagstamon git repo and made the following changes:

- Changed the version of pyqt5 in `build/requirements/macos.txt` to match the version installed by Homebrew (trying to install through pip failed, which is why I used the Homebrew package), in this case 5.15.4, which works, despite the note saying this version isn't working
- Changed the path for `pyinstaller` in `build/build.py` to be the correct one for arm64 Homebrew installations: `/opt/homebrew/bin/pyinstaller`

From there I did the normal steps to produce a macOS .app build:

- `pip3 install -r build/requirements/macos.txt`
- `cd build/ && python3 build.py`

Apple Silicon Macs require all arm64 code to be signed, but pyinstaller takes care of that by self-signing the .app bundle. For distribution, proper code signing would need to be set up, or as a worst-case, instructions given for how users could self-sign the app themselves.

Hope this is helpful to someone!

HenriWahl:
Thanks for your investigations.

The need for signed arm64 packages sounds alarming - on the one hand a good idea for commercial distribution but on the other I am not sure how some open source project shall handle this.

dotsam:
One other note here, RE: signing. The current build script which modifies the Info.plist after PyInstaller has run will not work with the out-of-box signing capabilities of PyInstaller, as editing the Info.plist file after signing will invalidate the signature. (This doesn't actually seem to be a problem for ad-hoc/self-signed apps, but would be for a properly signed .app for distribution).

[...]

PR #761, opened 2021-08-19
Build macos release from specfile

dotsam:
Per discussion in #743, here is a PR to use a predefined spec file with PyInstaller, which lets us specify extra plist keys and set a version within pyinstaller's system. This has the advantage of letting pyinstaller's automatic code signing system still work without invalidating the signature by modifying the Info.plist after it's been signed.

This also removes the hardcoded path to the `pyinstaller` binary, which should make building from source for Apple ARM users a little simpler.


HenriWahl:
Thanks!

HenriWahl:
Can https://github.com/HenriWahl/Nagstamon/issues/771 be caused by this PR?

PR #772, opened 2021-09-10
Add support for Zabbix 5.4.0+ and move from urllib to requests

dizixagi:
- Support for Zabbix 5.4.0+ (API changes, multiple interfaces support)
- Persistent connections via requests (reduce load on Zabbix www server)
- No longer using thirdparty.zabbix_api module
- Code cleanup

### DIFF
--- a/Nagstamon/Servers/ZabbixProblemBased.py
+++ b/Nagstamon/Servers/ZabbixProblemBased.py
@@ -4,7 +4,7 @@ import sys
 import time
 import logging
 import requests
-from packaging.version import parse as parse_version
+from pkg_resources import parse_version
 
 from Nagstamon.Helpers import HumanReadableDurationFromTimestamp
 from Nagstamon.Config import conf
--- a/Nagstamon/Servers/ZabbixProblemBased.py
+++ b/Nagstamon/Servers/ZabbixProblemBased.py
@@ -1,32 +1,104 @@
-# -*- encoding: utf-8; py-indent-offset: 4 -*-
-#
-# Zabbix.py based on Checkmk Multisite.py
+# encoding: utf-8
 
 import sys
-import urllib.parse
 import time
 import logging
-import datetime
+import requests
+from packaging.version import parse as parse_version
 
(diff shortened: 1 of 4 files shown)


HenriWahl:
Is the packaging python module necessary? It does at least not exist as RPM package in Fedora so it would be better to avoid it.
Isn't it by the way an option to use the already existing FetchURL() method which all the other servers are using?

dizixagi:
Removed "packaging" and used built in already present "setuptools" module named "pkg_resources".

I've looked at FetchURL() method but sadly it uses local instance of requests.Session() that gets deleted each time method is called. That breaks persistent connection and does not reduce load on WWW/API server.

HenriWahl:
> Removed "packaging" and used built in already present "setuptools" module named "pkg_resources".

Thanks - removes some dependency.

> I've looked at FetchURL() method but sadly it uses local instance of requests.Session() that gets deleted each time method is called. That breaks persistent connection and does not reduce load on WWW/API server.

Really? This is bad. Waht kind of a 'session' is this? Might it be somewhat possible to keep it?
The advantage of using FetchURL like the others would be same proxy settings and so on.

dizixagi:
Zabbix does not use HTTP authorization, it uses token that needs to be send alongside each request, so when using FetchURL() we would need to pass no_auth=True. That end us in line 1488 of Generic.py. There we have initialization of temporary_session object that gets deleted after each use. That makes Nagstamon open new connection to web server each time it is called, wasting time on socket opening, SSL handshakes etc, no keep-alive is used.

I made small compassion on how using singe instance of requests.Session() makes it way faster and less expensive on server side: https://gist.github.com/dizixagi/b13a225d4916eb99ea2dd948ffb430d9

I could look into changing that part of FetchURL() to implement proper keep-alive then rewrite Zabbix implementation once again, but for now I think using what I propose here in this pull requests gets us to support Zabbix 5.4.0+ and reduces stress on web servers running Zabbix, especially with multiple users running it simultaneously.

Issue #773, opened 2021-09-15
Icinga: When service.status_information is null we get an exception


glesys-andreas:
When the status_information is null from an Icinga server we get an error message:
"AttributeError: 'NoneType' object har no attribute 'replace'
This causes nagstamon not to display any checks

https://github.com/HenriWahl/Nagstamon/blob/aeef908d42a3d740be9ca9d8ce9f838a11050360/Nagstamon/Servers/Icinga.py#L197
This can be solved by adding an if statement before the variable assignation:
`                        if s['status_information']:
                            self.new_hosts[host_name].services[service_name].status_information = s['status_information'].replace('\n', ' ').strip()`

HenriWahl:
Could you provide a pull request, so this fix can be integrated faster?

glesys-andreas:
Created the pull request #774 , please let me know if i need more changes to the PR

HenriWahl:
Thanks. Will be available in some minutes in the latest releases.

PR #774, opened 2021-09-15
Fix issue when status_information is null

glesys-andreas:
Fix for issue: https://github.com/HenriWahl/Nagstamon/issues/773
When status_information is null, an exception is thrown and no lines are displayed


HenriWahl:
Thanks

PR #781, opened 2021-11-11
Allow to use display_name on services with Thruk

fufroma:
Hi,

I'm not fluent in Python, so it's not a pretty patch, but it's something ¯\_(ツ)_/¯

I use Thruk and display_name on services. I saw that it was enabled on Icinga, so I tried to do the same for Thruk.

Know issues:
- I didn't do the display_name for Host, only for services
- ~~on services with specific display_name, only the Monitor link work~~

Best regards,


HenriWahl:
Hi,
thanks for contributing. Give me a hint when it is ready for merge. :+1: 

Regards

fufroma:
Look good to me, everything seems to work. I checked Debian and Windows.

I think it's ready to merge :)

PR #796, opened 2022-01-13
Experimental Solarwinds Orion support

SKSEffertz:
This pull request adds experimental support for the Solarwinds Orion platform.  It connects to the Rest API using the Solarwinds Orion SDK and reads all active alerts. It does not use node status information as they cannot be acknowledged (but could be added if required)

### DIFF
--- a/Nagstamon/Servers/Orion.py
+++ b/Nagstamon/Servers/Orion.py
@@ -1,0 +1,433 @@
+# encoding: utf-8
+
+# Nagstamon - Nagios status monitor for your desktop
+# Copyright (C) 2008-2014 Henri Wahl <h.wahl@ifw-dresden.de> et al.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


HenriWahl:
Hi,
I read in Orion.py `# * Install the Orion python SDK - https://github.com/solarwinds/orionsdk-python` - what does it mean? Is this an external requirement to run it or can it be bundled?

SKSEffertz:
Hi,
the SDK is a python packages to wrap the REST API calls. It must be installed to run and compile nagstamon.py. For the bundled ("compiled") version the SDK does not have to be installed on the PC.

But you are right to mention the SDK. The license for the SDK is apache-2.0. So I think it should be compatible with the GNU GPL used by Nagstamon?

HenriWahl:
Well, at https://www.gnu.org/licenses/license-list.en.html#apache2 they state

> 
> [Apache License, Version 2.0](https://directory.fsf.org/wiki/License:Apache2.0) ([#apache2](https://www.gnu.org/licenses/license-list.en.html#apache2))
> 
> This is a free software license, compatible with version 3 of the GNU GPL.
> 
> Please note that this license is not compatible with GPL version 2, because it has some requirements that are not in that GPL version. These include certain patent termination and indemnification provisions. The patent termination provision is a good thing, which is why we recommend the Apache 2.0 license for substantial programs over other lax permissive licenses.
> 

Thus I fear it won't be possible to distribute it. To use it privately in your case shall be no problem. Maybe you could build yourself a binary for distribution in your organization?

SKSEffertz:
Hi Henri,

thanks for looking into the details. I already built the application for our internal use. It's working fine ;) My intention was to share the solution with the community.

Maybe Nagstamon will be GPLv3 licensed in the future. Until then I close the pull request.

Nagstamon is a great tool. Thank you for all the effort you put in it!

Best,
Stephan

HenriWahl:
Thanks anyway!

PR #801, opened 2022-02-09
Add feature: Zabbix Trigger Names and Improvements for Version higher 5.2

Simkimdm:
Hi, 
First of all thank you for developing this useful tool. Already using it since 8 years already.

I had the issue that Macros in Trigger names aren´t resolved in the Zabbix Integration. So i rewrote the Zabbix Script.

While make this possible, i also realize that some Changes coming with Zabbix 5.2 are moved out on Host level.
So it is not necessary to iterate over all hosts anymore, which reduce the load on the API in big Environments like ours with around 10000 hosts.
Since Version 5.4 there are less informations on Host API. But you can gather most of the information by using just the trigger API. So there is space for more improvements.

it ist just tested it for Version 5.4.

For using Maintenance you have to add the TAG: 'triggerid: {TRIGGER.ID}' for each Trigger. 
Through this you just set this specific Trigger in Maintenance.

So this is my first PR please let me know if there is something missing.
BR

### DIFF
--- a/Nagstamon/Servers/Zabbix.py
+++ b/Nagstamon/Servers/Zabbix.py
@@ -139,6 +139,7 @@ class ZabbixServer(GenericServer):
                                                        "jmx_available", "jmx_error", "jmx_errors_from",
                                                        "maintenance_status", "maintenance_from"],
                                             "selectInterfaces": ["ip"],
+                                            "with_triggers": True,
                                             "filter": {}
                                             })
             except (ZabbixError, ZabbixAPIException, APITimeout, Already_Exists):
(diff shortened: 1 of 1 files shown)


Nagstamon/Servers/Zabbix.py
```diff
@@ -472,98 +466,116 @@ def set_recheck(self, info_dict):
...
+            # 16 - unacknowledge event
+            actions = 2
+            # If sticky is set then close only current event
+            # if triggerid == service and sticky:
+            #     # do not send the "Close" flag if this event does not allow manual closing
+            #     triggers = self.zapi.trigger.get({
+            #         'output': ['triggerid', 'manual_close'],
+            #         'filter': {'triggerid': triggerid}})
+            #     if not triggers or 'manual_close' not in triggers[0] or str(triggers[0]['manual_close']) == '1':
+            #         actions |= 1
```
hunsbea, line 523: This breaks the ability to close problems ...

hunsbea:
I'm taking a look at the latest version of Nagstamon Zabbix code. I see somehow nearly all the services/alarms are duplicated now. 
image
I will do some additional testing (version 5.4) and probably raise another PR to fix anything else broken.

PR #809, opened 2022-05-10
Fix : Remove line returns from a Host status information

octvcdrc:
Hello,
This fix replace line returns with a space from a Centreon Host status information (exactly as it is done with status information)

This prevents this behavior :
before

After the fix : 
after


HenriWahl:
Thanks!

PR #816, opened 2022-06-24
Added SMHub monitor

delu86:
Hi,
I've added the support for SMHUB, a SAP monitor developed by Retelit.
https://smhub.retelit.it/

### DIFF
--- a/Nagstamon/QUI/__init__.py
+++ b/Nagstamon/QUI/__init__.py
@@ -3572,11 +3572,6 @@ class TreeView(QTreeView):
                 # action to signalmapper
                 self.signalmapper_action_menu.setMapping(action_menuentry, a)
                 action_menuentry.triggered.connect(self.signalmapper_action_menu.map)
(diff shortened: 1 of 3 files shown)


Nagstamon/QUI/__init__.py
```diff
@@ -3568,6 +3572,12 @@ def cell_clicked(self):
                 # action to signalmapper
                 self.signalmapper_action_menu.setMapping(action_menuentry, a)
                 action_menuentry.triggered.connect(self.signalmapper_action_menu.map)
+            SMHUB_USLESS_ACTION = ['Telnet','RDP','SSH','VNC']
```
HenriWahl, line 3575: These actions can be deleted by user anyway - I think they should not be handled here specifically for SMHub.

delu86: I have removed it in the last commit

HenriWahl:
Solved by https://github.com/manuelbolzoni/SMHstamon

PR #828, opened 2022-08-03
Fix zabbix bugs regarding timeout, ack/closing, and duplicate services

hunsbea:
1. deleted some commented code
2. In the Generic.py file, the below deleted code is redundant, and was causing Zabbix (and likely other monitoring systems) to display duplicate services
3. I now pass the "timeout" configured in settings to the ZabbixAPI. This fixes an issue to allow configuring a timeout longer than 10 seconds, which is required if your Zabbix instance is rather large.
4. Fixed a regression bug that removed the ability to "close" problems from Nagstamon


hunsbea:
Feel free to ask me any questions about this if you need clarification.

HenriWahl:
Thanks!

PR #835, opened 2022-08-17
 Fix centreonAPI JSONDecodeError after getting a token

Alyx-LeGuen:
Replay the request after getting a new token, otherwise result.result will be empty and it will throw an error "Error : json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)"


Alyx-LeGuen:
Sorry, I committed on the wrong line for my fix

PR #841, opened 2022-08-26
Fixed mutual default argument in methods.

nautics889:
Got rid from a mutual default argument `all_services` in components for
server interaction.
Mutual default argument (like list in `_set_acknowledge(...)` method) can cause unpredictable behavoiur.
```
>>> def foobar(bar=[]):
...     bar.append('test')
...     print(f'List {bar} with ID {id(bar)}')
>>>
>>>
>>> foobar()
List ['test'] with ID 140147921265664
>>> foobar()
List ['test', 'test'] with ID 140147921265664  # as you see function hasn't create an empty list during second call, instead picked already existed
```
You can see more at the following topics:
https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument
https://web.archive.org/web/20200221224620id_/http://effbot.org/zone/default-values.htm

Also you can simply check a list (or any different collection) for containing items like this:
```
bar = []
if bar:
    ...
```
instead of
```
bar = []
if len(bar) > 0:
    ...
```

### DIFF
--- a/Nagstamon/Servers/Alertmanager/alertmanagerserver.py
+++ b/Nagstamon/Servers/Alertmanager/alertmanagerserver.py
@@ -328,7 +328,7 @@ class AlertmanagerServer(GenericServer):
 
 
     def _set_acknowledge(self, host, service, author, comment, sticky, notify, persistent,
-                         all_services=[], expire_time=None):
+                         all_services=None, expire_time=None):
         alert = self.hosts[host].services[service]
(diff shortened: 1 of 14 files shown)


HenriWahl:
Sorry für the later answer - I did not find the time to check your PR but will do ASAP.

HenriWahl:
Thanks!

PR #852, opened 2022-10-08
Github workflows - replacing deprecated macos-10.15

Tontonitch:
Github workflows - replacing deprecated macos-10.15 by macos-12

### DIFF
--- a/.github/workflows/build-release-latest.yml
+++ b/.github/workflows/build-release-latest.yml
@@ -117,7 +117,7 @@ jobs:
           if-no-files-found: error
 
   macos:
-    runs-on: macos-10.15
+    runs-on: macos-12
     needs: test
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Tontonitch:
I had this message 4days ago when I tried to compile one of my test version of nagstamon:
[macos](https://github.com/Tontonitch/Nagstamon/actions/runs/3207020567/jobs/5241593989)
This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on December 1st, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583
[macos](https://github.com/Tontonitch/Nagstamon/actions/runs/3207020567/jobs/5241593989)
GitHub Actions has encountered an internal error when running your job.

That's why I thought that it was not possible anymore to use macOS-10.15

If it is still supported you can close that PR ;-)

HenriWahl:
If we get a stable 3.10 until December 1st there is a chance that some users of older Macs still can use the stable version of Nagstamon. After this date it make take time anyway until the next stable release.

PR #869, opened 2022-11-05
Add Values to ARP

Trenly:
When the package is installed, values for DisplayVersion and Publisher are not set in the registry. In order for other applications like [winget](microsoft/winget-cli) to properly interact with the application, these values should be set.

Current Behavior:
```
Starting package install...
Successfully installed
--> Refreshing environment variables
--> Comparing ARP Entries

DisplayName      DisplayVersion Publisher ProductCode
-----------      -------------- --------- -----------
Nagstamon 3.10.1                          {44F7CFFB-4776-4DA4-9930-A07178069517}_is1
```

New Behavior:
```
Starting package install...
Successfully installed
--> Refreshing environment variables
--> Comparing ARP Entries

DisplayName      DisplayVersion Publisher  ProductCode
---------------- -------------- ---------- -----------
Nagstamon 3.10.1 3.10.1         Henri Wahl {44F7CFFB-4776-4DA4-9930-A07178069517}_is1
```

### DIFF
--- a/build/windows/nagstamon.iss
+++ b/build/windows/nagstamon.iss
@@ -1,6 +1,8 @@
 [Setup]
 AppName=Nagstamon
 AppVerName=Nagstamon {#version}
+AppVersion={#version}
+AppPublisher=Henri Wahl
 DefaultDirName={commonpf}\Nagstamon
 DefaultGroupName=Nagstamon
 AlwaysUsePersonalGroup=false


HenriWahl:
Thanks!

HenriWahl:
@Trenly BTW: how do you use winget? I tried without success. Any prerequisites necessary?

Trenly:
> @Trenly BTW: how do you use winget? I tried without success. Any prerequisites necessary?

Depending on your version of Windows, you may already have it installed. If not, you can install `App Installer` from the Microsoft Store, or grab the latest release from [winget-cli](/microsoft/winget-cli/releases/latest), download the msixbundle, and install it. After installing it, load a new terminal and you should be able to use `winget -v` to check that it was installed successfully. For checking if you have any applications available to upgrade - `winget upgrade`, and there are similar commands for search / install / uninstall.

PR #882, opened 2022-12-02
Fallback from requests_kerberos to requests_gssapi

robert-scheck:
RHEL/CentOS 7 has no requests_kerberos, only requests_gssapi. This patch should keep the current behaviour, just adds the fallback additionally.


robert-scheck:
Alternatively, if you dislike approach, could we at least switch the dependencies, please? So use requests_gssapi as fallback if requests_kerberos is not available.

PR #883, opened 2022-12-02
Handle absence of attribute 'screenAt' at 'QApplication'

robert-scheck:
`QApplication` (using the default Qt5 and its default Qt5 Python 3 binding on RHEL/CentOS 7) has no attribute `screenAt`.


HenriWahl:
Thanks! :+1:

PR #886, opened 2022-12-04
Add Nagstamon RPM packages for RHEL 9 (and derivatives/clones)

robert-scheck:
This is a first attempt to add Nagstamon RPM packages for RHEL 9 (and derivatives/clones) - while hoping that I didn't get tricked somewhere during building and tests.

However this pull request is incomplete, because I don't know how to handle the `repo-fedora` step in `.github/workflows/build-release-{latest,stable}.yml` properly.

### DIFF
--- a/.github/workflows/build-release-latest.yml
+++ b/.github/workflows/build-release-latest.yml
@@ -138,6 +138,19 @@ jobs:
           retention-days: 1
           if-no-files-found: error
 
+  rhel-9:
+    runs-on: ubuntu-latest
+    needs: test
+    steps:
+      - uses: actions/checkout@v3
+      - run: /usr/bin/docker build -t build-nagstamon -f build/docker/Dockerfile-${{ github.job }} .
+      - run: /usr/bin/docker run -v ${{ github.workspace }}:/nagstamon build-nagstamon
+      - uses: actions/upload-artifact@v3
+        with:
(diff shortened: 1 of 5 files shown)


HenriWahl:
Thanks! I will prepare the repo so RHEL packages might be included as well.

HenriWahl:
I added extended the build pipelines to add RHEL to the repo on nagstamon.de. Can you pleae check if adding https://nagstamon.de/repo/rhel/nagstamon-latest.repo works for you?

robert-scheck:
For some reason, the RPM from the repository depends on python3-crypto and python3-pyqt6, which both shouldn't be the case.

First is unmaintained software according to https://src.fedoraproject.org/rpms/python-crypto, second is not available, because it is too new. But from my point of view, https://github.com/HenriWahl/Nagstamon/blob/master/build/docker/Dockerfile-rhel-9 should fit. Where is my mistake?

HenriWahl:
After the changes in https://github.com/HenriWahl/Nagstamon/commit/1f2cfcbdba0ebfaa0f67065c4732ca8fe1f5bdc9 I got it at least to install due to correct dependencies.

robert-scheck:
Ah, thank you! And yes, the package works for me as expected on Rocky Linux 9 here.

Issue #881, opened 2022-11-28
Nagstamon 3.10.1 doesnt work with Centreon version 21.x


Leon30071980:
Hi all, someone solve the problem with Centreon versione 21.x ? Other versions work fine.

"Unknown Error" is given with the 21.x version as imagine:
![image](https://user-images.githubusercontent.com/119290249/204230374-ef638dbe-2fb9-4977-b65b-8223ee80239e.png)

Can someone help us?
Or if the problem is known, can you please give me the link of the thread? 
Danke

Leon30071980:
Here is the debug log:

DEBUG: 2022-11-29 09:23:03.148315 Saving configuration to file c:\users\cdmmon\.nagstamon\nagstamon.conf
DEBUG: 2022-11-29 09:23:03.597324 Socage FetchURL: https://151.14.34.252/centreon/api/latest/platform/versions CGI Data: None
DEBUG: 2022-11-29 09:23:03.757618 [Socage] Status code 200
DEBUG: 2022-11-29 09:23:03.757618 [Socage] Loading class API, Centreon version : 21.4
DEBUG: 2022-11-29 09:23:03.757618 Socage Created server.
DEBUG: Initializing SystemTrayIcon
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE53F0> for state "OK"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE5360> for state "INFORMATION"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE5480> for state "UNKNOWN"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE5510> for state "WARNING"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE55A0> for state "AVERAGE"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE5630> for state "HIGH"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001D8D8FE56C0> for state "CRITICAL"
[...]

PR #894, opened 2023-01-20
[Centreon] Fix auth on 21.10

BenoitPoulet:
API v2 in 21.10 throw an error 500 when you have no valid token.

This fix : #881


HenriWahl:
Thanks @BenoitPoulet !

PR #895, opened 2023-01-26
This commit fixes the 'Monitor' links for Opsview.

johanthoren:
Previously, these links went to `/extinfo.cgi?/...` which is not correct and would result in a 404. I've replaced the links for hosts and services to instead use `/monitoring/#!?` with `autoSelectHost` and `autoSelectService` respectively.

Clicking 'Monitor' on a host or service in the tablewidget context menu now works.


HenriWahl:
Thanks @johanthoren !

PR #896, opened 2023-01-26
Add support for Opsview hashtag filtering

johanthoren:
- Add a new optional configuration field for Opsview hashtags.
- The configuration field takes a comma-separated list of hashtags.
- This will filter all objects to only display those with the selected tags.


HenriWahl:
Thanks @johanthoren !

PR #905, opened 2023-02-03
Add the can_change_only option for Opsview.

johanthoren:
This commit adds a new checkbox for Opsview servers that will filter out all services that the user cannot either change or set downtimes on.

It also contains some fixes where the `if conf.debug_mode` logic was missing from some Debug calls.

Also, it fixes a bug where clicking the link of the Opsview server address in the 'right click menu' would throw an error and crash the application.


HenriWahl:
Thanks!

PR #909, opened 2023-02-09
register nagstamon version on windows when installing via setup

### LINKED ISSUE #902: Installed Nagstamon version missing via winget (windows platform)
Hello,
on microsoft windows (e.g. win11) you can install and update programs via winget on command line.
With Nagstamon there seems to be an issue with the installed version.
Winget can see the new version but is unable to detect the installed version.
I tested with 3.10.1-win64 and the output of "winget upgrade --include-unknown" looks like:

Name             ID                  Version Source
-------------------------------------------------------------
Nagstamon 3.10.1 HenriWahl.Nagstamon Unknown 3.10.1    winget

Maybe the installer doesn't register the version properly(?)
The current version 3.10.1 is known but the installed version column shows "unknown".
Uninstall and fresh re-install shows the same behavior.

If further information is needed, just let me know.

Thanks! Thomas

thomass4t:
Registering the correct nagstamon version on windows is needed for handling updates and upgrades e.g. via winget.

Ticket-Nr: 902
Resolves #902

### DIFF
--- a/build/windows/nagstamon.iss
+++ b/build/windows/nagstamon.iss
@@ -1,7 +1,7 @@
 [Setup]
 AppName=Nagstamon
-AppVerName=Nagstamon {#version}
-AppVersion={#version}
+AppVerName=Nagstamon {#version_is}
+AppVersion={#version_is}
 AppPublisher=Henri Wahl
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks!

PR #929, opened 2023-04-12
Fix: minor updates in Monitos4x.py module

nautics889:
Remove unused variable `form_data` (the variable was declared redundantly).
So, supposing you have the code like this:
```
1 bar = []
2 if <some_condition>:
3     bar = json.dupms( ... )
4 else:
5     bar = json.dumps( ... )
6 print(bar)
```
...there's no need to create `bar` variable at 1st line, since it'd be defined on 6th line anyway.

___

Rename the variable `type` to `type_` since the name `type` shadows name for the built-in function `type()` which isn't a good practice.

___

p. s. the diff can seem quite big, pardon, that's just because mentioned variables appear all over the module

### DIFF
--- a/Nagstamon/Servers/Monitos4x.py
+++ b/Nagstamon/Servers/Monitos4x.py
@@ -164,8 +164,6 @@ class Monitos4xServer(GenericServer):
 
         # hosts
         try:
-            form_data = dict()
-
             page = 1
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks! 👍

PR #933, opened 2023-04-29
Fix: minor fixes; code styling

nautics889:
Simplified condition checking in several modules in Nagstamon/Servers.


nautics889:
This PR contains **only** code styling updates and not supposed to change logic or behavior.

HenriWahl:
@nautics889 so True... Thanks!

PR #950, opened 2023-07-13
fix 'NoneType is not subscriptable' error in Icinga2API

### LINKED ISSUE #949: Error "NoneType object is not subscriptable" using Icinga2Api
When connecting to a certain Icinga instance, it seems that some checks do not have a value for `last_check_result`. This causes the following error in Nagstamon, with no results displayed:

`TypeError: 'NoneType' object is not subscriptable`

sacredheartsc:
When connecting to a certain Icinga instance, it seems that some checks do not have a value for last_check_result. This causes the following error in Nagstamon, with no results displayed:

TypeError: 'NoneType' object is not subscriptable

Enabling debug mode revealed that the error was thrown from the line modified in this commit.

This commit wraps the assignment with a check for None to avoid this error.

fixes #949


HenriWahl:
Thanks!

PR #961, opened 2023-08-28
Stop alertmanager to fetch inhibited alerts

aleskxyz:
Default query for fetching alerts from alertmanager, fetches inhibited alerts also.
By the logic behind the inhibition, there is no need to fetch this alerts.
In this PR I have changed the inhibited query string param to false to filter out inhibited alerts.


aleskxyz:
Hi
No problem 
Inhibited alerts are alerts that are muted by other alerts. For example if cpu usage more that 80% is warning, and more than 90% is critical, when cpu usage goes above 90% both warning and critical alerts will fire.
In alertmanager you can inhibit warning alert by critical alert, so when you set inhibited=false, you will not get inhibited alerts in output.

PR #965, opened 2023-09-06
Zabbix requires downtime timestamps to be integers.

grzeg1:
May be some change in Python but mktime on newest Ubuntu's python returns fractional numbers.


HenriWahl:
@grzeg1 thanks!

Issue #966, opened 2023-09-21
Nagstamon-3.13-20230921 is broken with Centreon 22.10 (at least)


BenoitPoulet:
Hello,

A team mate installed the latest build and he is unable to load the status.

The problem was not present in the 3.10.

```
DEBUG: 2023-09-21 15:00:13.523119 Saving configuration to file c:\users\benoit.poulet\.nagstamon\nagstamon.conf
DEBUG: 2023-09-21 15:00:24.326904 Centreon Dev Created server.
DEBUG: 2023-09-21 15:00:24.326904 Centreon Pre Prod Created server.
DEBUG: 2023-09-21 15:00:24.326904 Centreon Prod FetchURL: https://centreon/centreon/api/latest/platform/versions CGI Data: None
DEBUG: 2023-09-21 15:00:24.549263 [Centreon Prod] Status code 200
DEBUG: 2023-09-21 15:00:24.549263 [Centreon Prod] Loading class API, Centreon version : 22.10
DEBUG: 2023-09-21 15:00:24.549263 Centreon Prod Created server.
DEBUG: 2023-09-21 15:00:24.549263 Demo Centreon User Created server.
DEBUG: 2023-09-21 15:00:24.549263 Demo Centreon admin Created server.
DEBUG: Initializing SystemTrayIcon
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD1FC0> for state "OK"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD1EA0> for state "INFORMATION"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD2050> for state "UNKNOWN"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD20E0> for state "WARNING"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD2170> for state "AVERAGE"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD2200> for state "HIGH"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x000001BEB0FD2290> for state "CRITICAL"
[...]

PR #967, opened 2023-09-26
Fixes for Nag. 3.13 and Centreon 22.10, 23.04+

BenoitPoulet:
Relative to #966

check_for_error() was not returning False anymore, but None.

Also fix some changes in the API starting from Centreon 23.04.

The case as been validated on Centreon 22.04 and 23.10beta


HenriWahl:
Thank you! :+1:

Issue #970, opened 2023-09-28
Error since IcingaDB-Web 1.1


Tontonitch:
Tested compatibility with the fresh new version of icingadb-web (https://github.com/Icinga/icingadb-web)

![image](https://github.com/HenriWahl/Nagstamon/assets/600025/367d0c8d-f27a-428e-a03d-54db8628586d)

Debug:
```
Traceback (most recent call last):
  File "Nagstamon\Servers\IcingaDBWeb.py", line 211, in _get_status
ValueError: could not convert string to float: '2023-09-28T20:50:02.582+00:00'
DEBUG: 2023-09-28 23:02:43.730600 xxxx FetchURL: https://xxxx/icingaweb2/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json CGI Data: None
ERROR: 2023-09-28 23:02:43.777469 xxxx Traceback (most recent call last):
  File "Nagstamon\Servers\IcingaDBWeb.py", line 211, in _get_status
ValueError: could not convert string to float: '2023-09-28T20:50:02.582+00:00'
```

tested with latest dev nagstamon 3.13 + 3.10.1.

Tontonitch:
Duplicate of #962

Issue #962, opened 2023-08-28
date string issue


sebek72:
Having issues with the latest 20230808 build.

monitor type: icingaDBWEB
backend: 
icingaweb2.11.4+ (build from master from August)
icingadb 1.1.0
icingadb-web (master, pre 1.1.0)

`DEBUG: 2023-08-28 12:37:00.660402 NEW PROD Start rechecking all
DEBUG: 2023-08-28 12:37:02.245729 NEW PROD Refreshing all hosts and services
DEBUG: 2023-08-28 12:37:02.876868 NEW PROD FetchURL: http://10.55.32.212:8080/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json CGI Data: None
ERROR: 2023-08-28 12:37:02.987765 NEW PROD Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/Nagstamon/Servers/IcingaDBWeb.py", line 211, in _get_status
    self.new_hosts[host_name].last_check = datetime.datetime.fromtimestamp(int(float(h['state']['last_update'])))
ValueError: could not convert string to float: '2023-08-25T07:02:14.894+00:00'`

P.S: older icingaweb2 (no icingadb) works ok.

HenriWahl:
@jr-timme any idea?

sebek72:
Hi. Thanks for the feedback. I would not know where to change that :)
I will try to install previous icinga stack versions to see if there is an instance where this works (btw my icinga is deployed via docker)

jr-timme:
Could you check the raw json output, just to make sure that the date there is already formatted?

```bash
curl -v -l -b cookies.txt -c cookies.txt "http://10.55.32.212:8080/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json" | jq
```

cookies can either be exported from a browser or generated through another curl in which you authenticate yourself

sebek72:
This is the json output:

`"last_update":"2023-09-21T09:49:38.753+00:00","last_state_change":"1970-01-01T00:00:00.000+00:00",
"next_check":"2023-09-21T09:49:59.838+00:00",
"next_update":"2023-09-21T09:50:21.846+00:00",`

mzac:
Just tired the update and still getting an error:

[...]

PR #978, opened 2023-10-30
fixed issue with date string in new icingadb version

DeadHunter:
As described in Issues #962 and #970, in the newer IcingaDB versions, last_update is returned as a string instead of a Unix timestamp.

The fix should also be backwards compatible.


HenriWahl:
Thanks!

PR #979, opened 2023-10-30
Refactor return value of check_for_error to None for all servers

aleskxyz:
check_for_error function returns None but some servers are expecting boolean.
I refactor all servers to check the return value explicitly is not None

### DIFF
--- a/Nagstamon/Servers/Alertmanager/alertmanagerserver.py
+++ b/Nagstamon/Servers/Alertmanager/alertmanagerserver.py
@@ -213,7 +213,7 @@ class AlertmanagerServer(GenericServer):
 
             # check if any error occured
             errors_occured = self.check_for_error(data, error, status_code)
-            if errors_occured is not False:
+            if errors_occured is not None:
                 return errors_occured
 
             for alert in data:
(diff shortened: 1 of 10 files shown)


aleskxyz:
I think you need to decide about the return value of this function.
In this PR you are checking that is it None or not!
https://github.com/HenriWahl/Nagstamon/pull/967

HenriWahl:
@aleskxyz thanks for the quick look for the cause, but in most cases the condition gets quite an opposite result now:

From

```python
if errors_occured:
```
to
```python
if errors_occured is not None:
```
there is an important difference, because in the first case it will only return if True, which extends to False in the second variant. I will have to check the consequences if done globally. Maybe a s a fix it will be enough if done in Prometheus and Alertmanager.

aleskxyz:
> @aleskxyz thanks for the quick look for the cause, but in most cases the condition gets quite an opposite result now:
> 
> From
> 
> ```python
> if errors_occured:
> ```
> 
> to
> 
> ```python
> if errors_occured is not None:
> ```
> 
> there is an important difference, because in the first case it will only return if True, which extends to False in the second variant. I will have to check the consequences if done globally. Maybe as a fix it will be enough if done in Prometheus and Alertmanager.

@HenriWahl 
I can use the first form but it seems we need to revert this PR also: https://github.com/HenriWahl/Nagstamon/pull/967

HenriWahl:
OK, I will look for it tomorrow.

HenriWahl:
@aleskxyz I checked it looks OK. Thanks for investigating!

HenriWahl:
It's included now in latest release 2.13-20231031.

HenriWahl:
Well, the whole code needs some refactoring...

PR #982, opened 2023-11-13
Fixing 4d3a851 that was using isnumeric on float

Tontonitch:
Fixing 4d3a851 that was using isnumeric on float.
Tested and works on both previous and new version of icinga2/icingadb (https://github.com/Tontonitch/Nagstamon/releases/tag/latest). 

However, there might be still a timezone problem, since with the new icingadb version the durations are calculated without considering that the dates returned by icinga2 are UTC. I need to recheck.

But at least now the build works on all icingadb versions :-)


HenriWahl:
Thanks!

PR #985, opened 2023-11-14
Fixing timezone issue when using latest version of icingadb

Tontonitch:
Before that fix, the utc tz was applied to the tz unaware datetime.now(), leading to incorrect duration calculations. Now fixed by retrieving local tz for the date retrieved by now(). Fix for python 3.3+ for better compatibility.
Tested and I have now the correct durations.

@HenriWahl now the compatibility with older and latest icingadb version should be complete ;-)


HenriWahl:
Thanks!

PR #987, opened 2023-11-15
Add map to down for Alertmanager

tk422:
This should allow mapping alert severity to the DOWN status in Nagastmon.

I think I got everything here but this is the first time I have played with this code


HenriWahl:
Thanks!

tk422:
@HenriWahl Was this merged before or after the latest release was built?

HenriWahl:
@tk422 it is part of the latest release 3.13-20231117.

Issue #219, opened 2016-04-11
[feature-request] Thruk, toggle backend


riskersen:
Hi all,

we're using Thruk as GUI for different Nagiose. In Thruk, you can toggle each Backend. 

It would be very nice to support this behavior in Nagstamon as well.

BR
Oliver

HenriWahl:
Nagstamon takes what it gets from your monitoring server and does not have any idea about any backend. Thruk presents its data in its way which is evaluated by the Nagstamon Thruk module so the backends should not have any influence.

riskersen:
So Thruk is not delivering the information from which Backend/Site the Alert is coming from?

HenriWahl:
I think so and otherwise see no use in this information if it even would be sent - the client will be interested in the state information, not how it was retrieved. If you choose Thruk as monitor type in Nagstamon it should get the right info from your Thruk server.

riskersen:
Lets imagine you're an administrator for several Nagios Systems of different Teams or Customers.

With Thruk you have a centralized overview of all Nagiose, but what if there are Nagiose you don't need to see everytime? In Thruk GUI, you see all configured backends/nagiose in the top bar which you can enable or disable. With that you'll get a better overview of what is happening for Nagiose where you actually need to do something on alerts.

When using nagstamon, its not possible. Either you have to configure aaall the Nagios GUIs from each server or you have to deal with a lot of information, which you don't need..


Do you get the point why i'm requesting something like that?

HenriWahl:
Your problem might be solved with using several monitors in one Nagstamon instance - see settings first "Servers" tab. These you can enable/disable as needed. Another solution is running several instances of Nagstamon with different configurations.

Any other solution might be pretty much overkill as not that many people use Thruk and of them not all will have your requirement - anyway I doubt that Thruk shows the underlaying monitoring server.

If anybody will code a working solution I won't hesitate to look at it but I am not able to provide one.

HenriWahl:
No real solution -> close

fermino:
Hey :)

I am currently needing this for my workplace for the same reasons stated above. We have a unified, multi-backend thruk instance which centralizes status information from different backends. Each area has its own backend for which they maintain the configuration, but we also have a dedicated monitoring team that should have access to a centralized dashboard with all the alerts. 

[...]

PR #991, opened 2023-11-28
Thruk: add disabled_backends field to filter out thruk backends (#219)

fermino:
This adds support for the feature asked in #219.

### DIFF
--- a/Nagstamon/Config.py
+++ b/Nagstamon/Config.py
@@ -977,6 +977,9 @@ class Server(object):
         self.notification_filter = "user.name=*"
         self.notification_lookback = "30 minutes"
 
+        # Thruk
+        self.disabled_backends = ""
+
 class Action(object):
     """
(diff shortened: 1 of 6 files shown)


fermino:
@HenriWahl I will need some feedback/advice as how to lay out the GUI elements (I hacked it together but pretty sure it could be better). Let me know if there's anything I can improve!

PS: Thanks for maintaining this project :)

HenriWahl:
I fixed the GUI a little - the result should be available soon at https://github.com/HenriWahl/Nagstamon/releases/tag/latest with release 3.13-20231130

fermino:
Sorry @HenriWahl wouldn't be necessary to merge thruk_backend into master? Do you want me to modify or test anything?

fermino:
@HenriWahl that definitely looks better. I would just make the label `Disabled backends (comma separated IDs)` or something similar, as it might not be that intuitive to use. Otherwise it looks great!

PR #994, opened 2023-12-08
harden #993 a bit more

tk422:
Hardens #993 
This adds a check to compare the labels (a dictionary) from each alert and if they match then dont create the new alert.  Havent seen it be an issue but should prevent an odd occassion where another alert comes through with the labels matching


HenriWahl:
:+1:

Issue #997, opened 2023-12-14
Converting Problem when using IcingaDB-Web


robertkrenn:
When swtitching the nagstamon Setting from Icingaweb2 to IcingaDB-Web, nagstamon only shows Error. The Debug Log shows further:

DEBUG: 2023-12-14 16:06:29.148170 Icingadb FetchURL: https://icinga2.augsburg.de/icingaweb2/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json CGI Data: None
ERROR: 2023-12-14 16:06:29.313691 Icingadb Traceback (most recent call last):
  File "Nagstamon\Servers\IcingaDBWeb.py", line 214, in _get_status
**ValueError: could not convert string to float: '2023-12-14T10:47:42.512+00:00'**

Nagstamon 3.12 is installed.

sebek72:
You need a newer version of nagstamon...take the latest build.

HenriWahl:
@robertkrenn which version of Nagstamon do you use? The testing releases contain IncingaDB-Web compatibility.

robertkrenn:
@HenriWahl We use the last stable Version 3.12.0, I'll try the latest build.

robertkrenn:
In Version 3.13- 20231208 I get the following error:


DEBUG: 2023-12-18 17:09:46.543998 Icingadb FetchURL: https://icinga2.augsburg.de/icingaweb2/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json CGI Data: None
DEBUG: 2023-12-18 17:09:46.626802 Icingadb FetchURL: https://icinga2.augsburg.de/icingaweb2/icingadb/hosts?host.state.is_problem=y&host.state.state_type=soft&columns=host.state.last_update&format=json CGI Data: None
DEBUG: 2023-12-18 17:09:46.693537 Icingadb FetchURL: https://icinga2.augsburg.de/icingaweb2/icingadb/services?service.state.is_problem=y&service.state.state_type=hard&columns=service.state.last_update,service.state.is_reachable&format=json CGI Data: None
ERROR: 2023-12-18 17:09:47.063121 Icingadb Traceback (most recent call last):
  File "Nagstamon\Servers\IcingaDBWeb.py", line 320, in _get_status
AttributeError: 'NoneType' object has no attribute 'replace'

Tontonitch:
[...]

Issue #996, opened 2023-12-11
IcingaDBWeb.py


Dymo-dcc:
IcingaDB setup error, tried with 3.12 and now again with 3.13

![Screenshot 2023-12-11 154715](https://github.com/HenriWahl/Nagstamon/assets/81287207/aaad0bf2-bdc2-47c5-aea7-fc5ba8cba62b)
![Screenshot 2023-12-11 154850](https://github.com/HenriWahl/Nagstamon/assets/81287207/81e1e478-b828-47b4-9f13-8ceff19d8b0d)
![Screenshot 2023-12-11 154818](https://github.com/HenriWahl/Nagstamon/assets/81287207/93a5b7eb-d624-4784-b7c2-e60f8cdb2e79)

*********************************

DEBUG: 2023-12-11 15:43:56.487680 Saving configuration to file c:\users\xxxxxx\.nagstamon\nagstamon.conf
DEBUG: 2023-12-11 15:43:56.703236 IcingaDB Created server.
DEBUG: Initializing SystemTrayIcon
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D460310> for state "OK"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D4601F0> for state "INFORMATION"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D4603A0> for state "UNKNOWN"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D460430> for state "WARNING"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D4604C0> for state "AVERAGE"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D460550> for state "HIGH"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D4605E0> for state "CRITICAL"
DEBUG: SystemTrayIcon created icon <Nagstamon.QUI.QIconWithFilename object at 0x0000024C9D460670> for state "DISASTER"
[...]

PR #999, opened 2024-01-06
Handle null JSON output in Icinga DB Web checks

DeadHunter:
- Fixes issue where a check with no output in Icinga DB Web is marked as null in JSON, resulting in 'none' in code
- Resolves processing error by casting the value to a string

Solves: #996 #997 

## Before
image
## After
Empty output will be displayed as `None` in the status information, instead of raising an error.
image


HenriWahl:
Thank you!

PR #1003, opened 2024-01-15
Fixing the use of isnumeric on float

Tontonitch:
Fixing the use of isnumeric on float. I thought I requested that change, but seems that it is not in the code.


Tontonitch:
For your information, without that change, the format of the dates reported by previous versions of icingadbweb is not well detected. This change fixes the problem.

Tontonitch:
Seems that #985 and #982 had been rolled back. I don't see since when, but the merged changes are not in master anymore. So both related problems are back.

HenriWahl:
@Tontonitch I found some differences which seem to got lost with merges. Can you please check latest master if it is now as expected?

Issue #872, opened 2022-11-09
Feature Request - Client Certificate


nunoseita:
First of all, thank you for an awsome product.

We use a CheckMK server that forces the use of client side certificate and only then we are presented with login prompt.
Is that feature possible to be implemented?

HenriWahl:
Because the underlaying Python module requests does not support client certificates which are protected by password (https://requests.readthedocs.io/en/latest/user/advanced/#client-side-certificates) until now this feature was considered not very useful.

nunoseita:
Maybe the feature could have the warning: `The private key to your local certificate must be unencrypted. This is a security flaw if the key itself is not in a encrypted filesystem or access to it is not controlled.`
And if the key is compromised we can allways revoke and issue a new cert for that client.
Can you consider to make this feature?

HenriWahl:
@nunoseita do you have the resources to create a prove of work for this? I cannot test this due to missing setup.

spacefreak86:
@HenriWahl since version 3.3, the SSLContext.load_cert_chain function has an optional argument password to decrypt the key. If the key is not encrypted, the argument is simply ignored: https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_cert_chain

So the client certificate authentication could be implemented in a secure way.

HenriWahl:
@spacefreak86 do you know how to integrate this into the used **requests** module?

HenriWahl:
@spacefreak86 Nagstamon uses https://requests.readthedocs.io in https://github.com/HenriWahl/Nagstamon/blob/master/Nagstamon/Servers/Generic.py#L32.

spacefreak86:
The requests module itself is able to handle client certificates: https://docs.python-requests.org/en/latest/user/advanced/#client-side-certificates

Unfortunately, it does not support using encrypted keys. In my opinion, private keys should never be stored unencrypted, so I suggest to use https://github.com/m-click/requests_pkcs12 . 
In addition to the better security, this is convenient for the users as well, as client certificates are usually stored in PKCS12 format.

What do you think about it?

HenriWahl:
@spacefreak86 this looks interesting.
Would you be able to try an implementation?

NotoriusNeo:
Hello!

I've tried using BeautifulSoup to simulate a browser in the terminal... 

```
from bs4 import BeautifulSoup
from requests import Session
from requests_pkcs12 import Pkcs12Adapter
[...]

PR #1022, opened 2024-04-04
Add client-side certificate support

alexharrington:
Add support for client-side certificates to nagstamon. I've only tested with nagios as I don't have access to others.

I've never used QT before and the UI side of this is not perfect - if you untick the option to use a client certificate then the form does not shrink back down as it should.

HenriWahl/Nagstamon#872

### DIFF
--- a/Nagstamon/Config.py
+++ b/Nagstamon/Config.py
@@ -122,7 +122,9 @@ CONFIG_STRINGS = ['custom_browser',
                   'autologin_key',
                   'custom_cert_ca_file',
                   'idp_ecp_endpoint',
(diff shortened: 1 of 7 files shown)


HenriWahl:
Hi @alexharrington,
sorry for the late reply. What client side platfprm did you test it on?

alexharrington:
I tested on Ubuntu 22.04 with Nagios Core 4.4.

nunoseita:
Hi @alexharrington @HenriWahl can you guys merge this into main branch?
This feature nowdays is so important.

PR #1043, opened 2024-06-13
Icinga2 API support improvements

brenard:
Hello,

This PR provide some improvements on the existing Icinga2 API support:

- remove dependency on `icinga2api` lib (project dead/inactive since 6 years)
- add menu actions

__Note:__ Implementation based on current [Icinga2 API documentation](https://icinga.com/docs/icinga-2/latest/doc/12-icinga2-api) and tested on Icinga2 version _2.13.6_ (from Debian stable).


HenriWahl:
Thanks!

PR #1046, opened 2024-07-08
Add compatibility for CheckMK 2.3

jkim2492:
Since the upgrade from version 2.2 to 2.3 of CheckMK, the downtime and recheck commands stopped working from Nagstamon (acknowledge is still working). It seems like CheckMK is trying to deprecate the legacy API in favor of the new REST API.

We haven't been able to pin-point the _exact_ change that caused this incompatibility. However, we were able to find evidences that CheckMK is trying to drop the legacy API in favor of the REST API. Below are some examples of changes that force users to use the REST API:
* [Werk #16689: Decommission legacy check API](https://checkmk.com/werk/16689)
> Commandline call plugins for special agents and active checks in this folder will still work, but we provide a new API for those as well now (see [Werk #16259](https://checkmk.com/werk/16259)). They will stop working in Checkmk 2.4.
* [Werk #13640: Deprecation of Web API](https://checkmk.com/werk/13640)
> We recommend migrating all existing scripts that use the Web API to the REST API as soon as possible to catch issues early before you move to CheckMK 2.2.

Changes made in this PR have been tested against:
* Legacy CheckMK `1.2.6p12` on Bionic
* OMD CheckMK `2.2.0p26.cre` on Focal 
* OMD CheckMK `2.3.0p3.cre` on Jammy

### DIFF
--- a/Nagstamon/Servers/Multisite.py
+++ b/Nagstamon/Servers/Multisite.py
@@ -25,6 +25,8 @@ import urllib.request, urllib.parse, urllib.error
 import time
 import copy
 import html
+from datetime import datetime
+from zoneinfo import ZoneInfo
 
 from Nagstamon.Objects import (GenericHost,
                                GenericService,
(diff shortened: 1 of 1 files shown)


HenriWahl:
So this pull request is now replaced by https://github.com/HenriWahl/Nagstamon/pull/1047 ?

jkim2492:
Yup. Sorry for the confusion

Issue #1070, opened 2024-10-22
Incorrect display of last_check field in Nagstamon


wigor2206:
A good day!

We use Nagstamon in conjunction with Icinga to quickly respond to critical inspection conditions.

The problem is that in Nagstamon in the last_check line the format of the output time is very large.
![image](https://github.com/user-attachments/assets/5448bde8-1ba8-40ca-92f3-ea39066cec83)

As example checked by user with UTC:
curl -k -u 'admin' -H 'Accept: application/json' -X GET '[https://monitoring-private.com/icingadb/services?service.name=A%20CPU%20Load&host.name=Jump-Host01.com'](https://monitoring-private.com/icingadb/services?service.name=A%20CPU%20Load&host.name=Jump-Host01.com%27)
_[
  {
    "id": "84b44d88d621abfef343c2b8890b841132e930b2",
...
      "last_update": "2024-10-18T08:59:11.717+00:00",
      "last_state_change": "2024-10-18T08:03:40.388+00:00",
      "next_check": "2024-10-18T09:03:59.999+00:00",
      "next_update": "2024-10-18T09:07:59.999+00:00",
...
      }
    },
    "host": {
      "id": "1cc6469c8aece4d43012a3554dcad55447b83395",
...
        "last_update": "2024-10-18T09:00:59.095+00:00",
        "last_state_change": "2024-10-10T11:09:59.637+00:00",
        "next_check": "2024-10-18T09:05:58.174+00:00",
        "next_update": "2024-10-18T09:10:59.145+00:00"
      }
    },
    "icon_image": {
      "id": null,
      "icon_image": null,
      "environment_id": null
    }
  }
]_

If you make a query from the database, the time format is Unixtime
_icingadb=# select * from service_state limit 1;
-[ RECORD 1 ]---------------+-------------------------------------------
id                          | \x9f6acc5d9c03070ef0d34baf54c905c09ca8968b
host_id                     | \x6a7972cd428aa690f18ecf0a65efe566d0fd08de
service_id                  | \x9f6acc5d9c03070ef0d34baf54c905c09ca8968b
environment_id              | \x2219331a9a0b312fd1ea3cb6d26111062c405792
properties_checksum         | \x98731a49fee07a22060194fe486a0bad744c18b9
state_type                  | hard
soft_state                  | 3
hard_state                  | 3
previous_soft_state         | 99
previous_hard_state         | 99
check_attempt               | 1
severity                    | 2112
output                      | 
[...]

PR #1075, opened 2024-10-25
Incorrect display of last_check field

wigor2206:
The problem with incorrect display of time in the last check line is described in the request https://github.com/HenriWahl/Nagstamon/issues/1070 .
We tried to fix a few lines in the IcingaDBWeb.py file to make the display of the time of the last validation check more compact.


HenriWahl:
@wigor2206  please check if latest testing build works as expected - thanks!

wigor2206:
@HenriWahl tested. Works. But maybe you need to think about not using the utc_tim variable, but how to make it more universal

wigor2206:
If it works out, I'll make a new PR

HenriWahl:
> @HenriWahl tested. Works. But maybe you need to think about not using the utc_tim variable, but how to make it more universal

What do you mean exactly?

wigor2206:
@HenriWahl 
Everything works as expected.
Should I use the latest branch or will there be a new release?

HenriWahl:
Better stick with the latest branch, I do not know when the next release will be in sight.

PR #1079, opened 2024-11-14
Update nagstamon.iss

Fvbor:
While building I recieved following warning:

Warning: Architecture identifier "x64" is deprecated. Substituting "x64os", but note that "x64compatible" is preferred in most cases. See the "Architecture Identifiers" topic in help file for more information.

https://jrsoftware.org/ishelp/index.php?topic=archidentifiers

After changing this line, the build runs without any problem


HenriWahl:
Thanks! :+1:

PR #1084, opened 2024-11-29
Update CentreonAPI.py

weeboo:
add hosts and services filters to the API request in order to reduce network traffic when there is large number of non ok services which your are not accoutable
I am not able to test the change on my laptop but if you compile a test version I will test.

### DIFF
--- a/Nagstamon/Servers/Centreon/CentreonAPI.py
+++ b/Nagstamon/Servers/Centreon/CentreonAPI.py
@@ -324,9 +324,9 @@ class CentreonServer(GenericServer):
         # replace )( by )|(
         self.re_service_filter = ''
         self.re_host_filter = ''
-        if conf.re_service_enabled is True and conf.re_service_reverse is False:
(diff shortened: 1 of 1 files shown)


BenoitPoulet:
Hello,

It’s not working for the Host filter, and the fields used for the search are not the good ones.

Must be 
s.description: search only in the name of the service (shown in the Resource column)

For the host I cant’t figure why it’s not working as the syntax seems ok.

HenriWahl:
@BenoitPoulet is this OK now?

BenoitPoulet:
The fields used for the search are now OK.

The filters seems to be OK, if we read "Regular expression for hosts"  as "Filter out with this regular expression for hosts".
So checkbox "revert" will means "Show me only hosts who match the regular expression for hosts"

But the patch is not optimized if the field means "Filter out with this regular expression for hosts", because we can ask directly to Centreon to do the job. Anyway Nagstamon actually do the filter out properly.

HenriWahl:
Thanks @weeboo and @BenoitPoulet

PR #1092, opened 2025-01-28
Fix for check mk2.3 downtime

brtdl:
Added fix for CheckMK 2.3. Previous versions were unable to set downtime (only tested on Windows) because of 'localtime' no longer allowed/identified as known timezone.


HenriWahl:
Thanks @brtdl I merged and cleaned up a little.

PR #1057, opened 2024-09-22
Add support for suppressed problems

dizixagi:
Add support for suppressed problems - Nagstamon will treat them as acknowledged. Feature introduced in Zabbix 6.2.0

PR #1098, opened 2025-02-16
Add support for suppressed problems v2 with version check fix

dizixagi:
Add support for suppressed problems - Nagstamon will treat them as acknowledged. Feature introduced in Zabbix 6.2.0.
Tested on Zabbix 7.0.x LTS version.

Fixed issue with #1057 being broken after #1091


HenriWahl:
@dizixagi thank you for you patience - should work now.

PR #1099, opened 2025-02-17
fix(alertmanager): Continue loop when severity is None

adrian-salas:
Hello,

When the alert is on severity: None, the next alerts are not displayed. It's probably due because the loop is exited with a break, but probably should continue to the next items

I propose this change from break to continue so the loops can iterate on all items


HenriWahl:
@adrian-salas thanks! In testing release now.

PR #1104, opened 2025-03-13
fix IcingaWeb2.py get_status uri

ng-pe:
Using Nagstamon with my icingaweb2 server behind a WAP, I've just noticed that the check URIs are missing the “=” for the check URIs in get_status...

So here's a little fix to correct the problem.


HenriWahl:
Thanks @ng-pe!
Please check latest testing release 3.17-20250319.

ng-pe:
Sorry for the delay in responding, 
3.17-20250319 did have the patch, but it is no longer available in the latest version 3.17-20250612.
Can you add it to the latest version?

PR #1122, opened 2025-07-08
fix_zabbix

Simkimdm:
Simplify Zabbix Integration. Move from urllib2 to requests. Raise Exception for older versions


HenriWahl:
Hey @Simkimdm - thanks for your fixes. Just one question: do you think it will be a big change to use the fetch_url() method most of the other server type use? See it at https://github.com/HenriWahl/Nagstamon/blob/3.17-20250612/Nagstamon/Servers/Generic.py#L1441 - this way all proxy and SSL settings would be used the same way in all servers. If this would mean too big changes, I might investigate this one day.

Simkimdm:
Hi @HenriWahl,
i´m on that.

HenriWahl:
@Simkimdm thank you. I will check when back from vacation.

PR #1135, opened 2025-08-27
Standardize qui module paths to align with new `QUI` directory structure

Simkimdm:
Hi @HenriWahl,
while debugging under Macos i got the Message: Module Nagstamon.qui not found
So I debugged it and there seems to be some issues regarding case sensitive folder structure, so changed all Nagstamon.qui -> Nagstamon.QUI which fixed my issue. it should also run under Windows ;)


HenriWahl:
I feared something like that on non Linux systems... actually I want to refactor QUI to qui to match best practice. As the macOS build runs too - how did you achieve that mixed state?

Simkimdm:
primary it doesnt run cause the folder upper case QUI. it tested it with small letters but Nagstamon.QUI is also mentioned a lot, so i thought the folder QUI -> so its the way to go ;)

Simkimdm:
Ah just rename the QUI folder to lowercase, works seemless. The remaining files were tempfiles from builds previously.

PR #1136, opened 2025-08-27
Zabbix Server fix auth and reload issues.

Simkimdm:
This fixed for me the crash when changes are made on Server Settings.
it also should fix the issue while authentication. Some errors from the community i can´t reproduce with my instances driving 7.0 & 7.4 Zabbix Instance docker & local hosted with different user settings (user/admin/superadmin)


HenriWahl:
Thanks! 🙏

PR #1145, opened 2025-10-02
Better error handling for Centreon

BenoitPoulet:
I tried to catch and return to the user more errors.
I improved the `__init__ ` to avoid Nagstamon to crash on errors. But I was not able to return information to the user at this step.


HenriWahl:
Thanks @BenoitPoulet!

PR #1154, opened 2025-11-19
Load existing cookies from db before opening web auth browser

Kleinrotti:
When re-authenticating with web auth no existing cookies get loaded like Idp login.
This change loads existing cookies from the db that the Idp cookie is present when re-authenticating and the user doesn't need to login at Idp again.

### DIFF
--- a/Nagstamon/qui/dialogs/weblogin.py
+++ b/Nagstamon/qui/dialogs/weblogin.py
@@ -15,11 +15,13 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA
 
+import datetime
 from Nagstamon.cookies import (cookie_data_to_jar,
                                load_cookies,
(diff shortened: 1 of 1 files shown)


HenriWahl:
I was just curious what Copilot would say and will review myself too, but what do you think of this review so far? The 'ascii' part is something I thought too at a first sight.

Kleinrotti:
I was also thinking about ascii or utf-8 when writing the code. RFC says only ascii are allowed in cookie names and values but I'm not sure if all websites follow that.

I would probably go with:
`qcookie.setName(cookie["name"].encode("ascii", errors="ignore"))
qcookie.setValue(cookie["value"].encode("ascii", errors="ignore"))`
to meet the standards.

Regards expiration date
`qcookie.setExpirationDate(datetime.datetime.fromtimestamp(cookie["expiration"]))`

setExpirationDate() can take QDatetime or datetime.

HenriWahl:
Thanks!

PR #1155, opened 2025-11-20
Fix IcingaDB Web 1.3 compatibility

brenard:
Following an update to Icinga, specifically IcingaDB Web to version 1.3, I noticed that no information was being displayed in Nagstamon. After investigating, it appears that the IcingaDB Web API now requires all relevant columns to be explicitly requested on getting status.

Additionally, I noticed that while the codebase supports the `expire_time` field, the UI did not provide a way to input this information. I fixed this and, in doing so, realized that the timezone of the submitted date was not being interpreted correctly by Icinga (resulting in a 1-hour offset in my case). After some research, I found that the `icingaweb2-tzo` cookie seemed to indicate the client's timezone, so I ensured it was included.

Finally, when running the application from the command line, I encountered annoying `MarkupResemblesLocatorWarning` warnings, which I suppressed using the [recommended method](https://git.launchpad.net/beautifulsoup/tree/bs4/_warnings.py#n26).

__Note:__ I don’t have an IcingaDB Web instance running version 1.2.x to verify compatibility, but since I only added fields to the query, I don’t expect it to cause any issues.


HenriWahl:
Thanks!

sasbibic:
Is an executable for windows x64,  that solves this problem available ?

HenriWahl:
@sasbibic check https://github.com/HenriWahl/Nagstamon/releases/tag/latest

sasbibic:
In this version, there is still only one property listed in columns, according to Log anyway ....

```Log
DEBUG: 2025-11-24 18:31:15.490992 3qit control fetch_url: http://my.icinga.server/icingaweb2/icingadb/hosts?host.state.is_problem=y&host.state.state_type=hard&columns=host.state.last_update&format=json CGI Data: None
ERROR: 2025-11-24 18:31:15.624807 3qit control Traceback (most recent call last):
  File "Nagstamon\servers\IcingaDBWeb.py", line 198, in _get_status
KeyError: 'display_name'
```

Regards
/Sas

xeiss:
@HenriWahl Commit d65a80d seems to be not included in Nagstamon-3.17-20251121. This version still has the same error "KeyError: 'display_name'" because it still asks only for "columns=host.state.last_update" (checked in Debug Log)
I also checked the Source Code Zip from the release "https://github.com/HenriWahl/Nagstamon/archive/refs/tags/latest.zip" and it is missing the changes from commit d65a80d

HenriWahl:
Apparently the commit was not applied on my branch which I fixed now. A new build is under way: https://github.com/HenriWahl/Nagstamon/actions/runs/19674624592

PR #1165, opened 2026-01-20
Fix/IcingaDBWeb: clean stale Icingaweb2 cookie only when Keycloak session is expired

### LINKED ISSUE #1163: Web login fails if old cookies.db exists
If I log into an IcingaWebDB server via web login using a Keycloak server without a cookies.db file, it works without any problems. As long as the Keycloak session is still valid, the login works even after restarting Nagstamon.

As soon as the sessions expire (e.g., after a day), web login via Keycloak fails – I'm always redirected to the login screen.

If I delete the cookies.db file containing the "old information," web login works again, and a fresh cookies.db file is created.

Information:
Nagstamon Version: 3.17-20251203

Monitored Server:

Monitor Type: IcingaDBWeb
Use Proxy: false
Ignore SSL/TLS certificate: true
Authentication: Web
Timeout: 10 seconds

Unfortunately, I don't see any relevant information in the log file from debug mode.

Plattenspatz:
Fixes #1163

In an IcingaDBWeb + Keycloak setup, Nagstamon can get stuck in a redirect-to-login loop when a persisted **Icingaweb2** session cookie from an older run is reused after the Keycloak session already expired server-side.

This PR adds a minimal, server-type-specific cleanup for IcingaDBWeb:
- run cleanup only once per application start / server instance
- only if Keycloak cookies are present
- only if there is no valid **KEYCLOAK_SESSION** cookie (based on the local expiration timestamp)
- then remove the persisted **Icingaweb2** cookie from **cookies.db**

Additionally, it guards **_set_recheck()** against missing HTML form elements to avoid a crash when authentication expired and the expected form cannot be found.

### DIFF
--- a/Nagstamon/cookies.py
+++ b/Nagstamon/cookies.py
@@ -183,24 +183,6 @@ def load_cookies():
             cookies[cookey]['value'] = decrypted_value.decode()
     return cookies
 
-def purge_cookie_by_name_if_session(server_name: str, cookie_name: str):
-    """
-    Delete a specific cookie for a server from SQLite DB if it is a session cookie (expiration IS NULL).
-    Intended to clean up persisted session cookies from older installations.
-    """
-    init_db()
-    connection = sqlite3.connect(COOKIE_DB_FILE_PATH)
-    cursor = connection.cursor()
-
-    cursor.execute(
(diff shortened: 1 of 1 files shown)


HenriWahl:
Thanks @Plattenspatz!

Issue #1144, opened 2025-09-30
Zabbix "monitor" link is broken


sylvainfaivre:
When you ckick an alert in Nagstamon, there's a "monitor" link in the popup menu.

This link is broken, redirects to an error page stating :  
```
Fatal error, please report to the Zabbix team

    Incorrect value for "hostids" field.
    Controller: problem.view
    action: problem.view
    filter_set: 1
    hostids: array
    show_suppressed: 1
```

I looked in the web server logs, and it seems the `hostids` parameter is missing a value. The URL generated by Nagstamon is `/zabbix.php?action=problem.view&hostids[]=&filter_set=1&show_suppressed=1`

HenriWahl:
@Simkimdm do you have any idea? 😇

HenriWahl:
@sylvainfaivre is this still an issue?

rayman93:
@HenriWahl seems to work in the latest development build

PR #1189, opened 2026-03-06
Fix for #1144 

Simkimdm:
prevent overwriting hosts, collect multiple alerts per host.


Simkimdm:
fix for #1144 and #1189

HenriWahl:
@Simkimdm thank you for the fast fix!

PR #1192, opened 2026-03-12
Add OIDC authentication support for IcingaDB Web

baralai329:
Adds support for OIDC (OpenID Connect) Authorization Code + PKCE authentication, enabling Nagstamon to connect to IcingaDB Web instances protected by an OIDC provider (tested with IcingaWeb2 v2.15.2 and Keycloak 26.0.5).

### DIFF
--- a/Nagstamon/config.py
+++ b/Nagstamon/config.py
@@ -120,7 +120,8 @@ CONFIG_STRINGS = ['custom_browser',
                   'autologin_key',
                   'custom_cert_ca_file',
                   'idp_ecp_endpoint',
-                  'monitor_site'
+                  'monitor_site',
+                  'oidc_client_id'
                   ]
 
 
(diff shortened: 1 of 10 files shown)


baralai329:
The pipeline failure is pre-existing and unrelated to my changes.

HenriWahl:
Howdy @baralai329 - what a big PR!

I have some remarks:
- do you think the OIDC functionality can be reached for all types of monitoring servers, not only IcingaWeb?
- I would first mature cookie based web login for all server types which means some clean up and unification in their code base before adding the next auth option, so I would like to delay this a little.

Best regards and thanks for contributing!

baralai329:
Howdy, thanks for the review!

That makes sense. I’ve gone ahead and implemented a more flexible solution that isn’t limited to IcingaWeb. You can check it out here: https://github.com/HenriWahl/Nagstamon/pull/1196

I am closing this PR in favor of the new flexible solution. Let me know what you think about it.

PR #1196, opened 2026-03-19
Add external auth helper support (OIDC/OAuth2, cookies, etc.)

baralai329:
Adds a generic auth helper interface that lets Nagstamon delegate authentication to an external command. This enables OIDC, OAuth2, SAML, or any custom auth scheme without modifying Nagstamon's core.

Providing documentation and an example implementation for others to use.

image

### DIFF
--- a/Nagstamon/config.py
+++ b/Nagstamon/config.py
@@ -120,7 +120,9 @@ CONFIG_STRINGS = ['custom_browser',
                   'autologin_key',
                   'custom_cert_ca_file',
                   'idp_ecp_endpoint',
-                  'monitor_site'
+                  'monitor_site',
+                  'auth_helper_command',
+                  'auth_helper_extra_args'
                   ]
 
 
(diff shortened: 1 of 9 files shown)


HenriWahl:
Thanks @baralai329 - I will need some time to check.

HenriWahl:
Just one question: can you explain how this is intended to work? What is the external auth helper?

baralai329:
In summary the the external auth helper is a standalone command-line program that you write and configure to handle authentication for Nagstamon.

When authentication is needed, Nagstamon calls the auth helper.

That external helper performs the actual login process (for example with OIDC or OAuth2) and needs to return headers (e.g. token) or a cookie in the specified format.

Nagstamon then uses those returned credentials to make authenticated requests to the systems it monitors. It does that by adding these credentials to the HTTP session.

HenriWahl:
Sorry for the late answer, but I am actually not convinced that this merge request should be merged. If the authentication is delegated to some not yet existing external script, what is the use of it? So the magic work needs to be done somewhere else. I think this makes it all more complicated.

baralai329:
I understand the concern about complexity. The reason the auth logic lives externally is that every organization's SSO setup is different (different providers, client IDs, token endpoints). It's the same pattern Git uses with git credential helpers and kubectl uses with --exec credential plugins. The core tool stays simple, and the environment-specific auth is handled outside.

For existing users, nothing will change. This is purely opt-in via a new config field. No one who doesn't configure it is affected.

Here's a concrete use case: I'm trying to monitor services protected by a corporate OIDC SSO. Nagstamon currently has no way to authenticate against these. This PR makes that possible without baking one specific OIDC implementation into the core.

lunachocken:
I can agree with @baralai329 here. Their approach would make Nagstamon not be required to support each and every use case, instead it should simply expect perhaps a token/JWT/etc

This would mean businesses can simply maintain/distribute a helper script/executable for internal use without requiring patching within Nagstamon to be able to authenticate with their own internal authentication.

-> Set external CLI tool e.g. https://github.com/SecureAuthCorp/oauth2c
-> External tool requests user logs in, via web/other method
-> Returns oauth token to Nagstamon
+ potentially validates the current one set is correct?

Like @baralai329 said, other tools already implement support via external authentication tools. An example is azure integrates into the kubectl command via an add-on. 

This means kubectl isn't platform dependent but rather is platform agnostic for auth, which is a better approach.

Friedrich-Franke:
I'd like to add a real-world use case in support of this, since I'm in exactly the situation it solves.

Our Icinga Web 2 instance sits behind a corporate OIDC single sign-on that is backed by certificate-based (smartcard) authentication. There is no local API user available to us, and the REST API port is firewalled off, so the only supported way in is the SSO browser flow. Nagstamon currently can't authenticate against that — `basic`, `digest` and `kerberos` all fail, because the session ultimately depends on an OIDC token that none of the built-in methods can obtain.

That is exactly why a generic helper interface makes more sense than baking one OIDC implementation into the core: every organization's SSO differs (provider, tenant, client ID, token endpoint, how the credential is acquired). It's the same reasoning behind `git credential` helpers, `kubectl --exec` credential plugins and Docker credential helpers — the tool stays simple and protocol-agnostic, while the environment-specific auth lives in a small external script that just returns a header or cookie.

Since it's opt-in via a single new config field, nobody who doesn't configure it is affected.

Happy to test this branch against a real OIDC/SSO-protected Icinga setup and report back, if that helps with the review.

HenriWahl:
Hi there - I will look at it but as always time is sparse so it might be still delayed.

Issue #1197, opened 2026-03-19
Hosts with UNKNOWN status are not shown in the interface


blaboucheShinken:
Hello,

In the GenericServer class, the variable self.nagitems_filtered does not contains the key for "UNKNOWN" host, and in the get_status() function the state "UNKNOWN" is not tested.
We changed this behaviour and added an option to filter out UNKNOWN hosts. A pull requests with our work is coming soon.

Best regards,
Bastien from Shinken Solutions

PR #1198, opened 2026-03-19
Hosts with UNKNOWN status are not shown in the interface

blaboucheShinken:
Hello,
 
Here is the pull request mentioned in the #1197
We noticed that Livestatus creates hosts with the UNKNOWN state and these hosts weren't displayed in the interface, because the variable self.nagitem_filtered didn't had an entry for that state for hosts, only for DOWN and UNREACHABLE states :

```
self.nagitems_filtered = {'services': {'DISASTER': [], 'CRITICAL': [], 'HIGH': [],
    'AVERAGE': [], 'WARNING': [], 'INFORMATION': [], 'UNKNOWN': []},
    'hosts': {'DOWN': [], 'UNREACHABLE': []}}

```

So we added it and we can now use it to store hosts with this state coming from Livestatus servers and display them in the interface. 
We also added the configuration option to filter out UNKNOWN hosts with the current filter system.
 
We are open to any comment, if you feel that some changes need to be made or if you would like things to be handled differently, feel free to let us know and we will adjust accordingly.
 
Best regards,
Bastien from Shinken Solutions

### DIFF
--- a/Nagstamon/config.py
+++ b/Nagstamon/config.py
@@ -171,6 +171,7 @@ class Config:
         self.filter_all_down_hosts = False
         self.filter_all_unreachable_hosts = False
         self.filter_all_unreachable_services = False
+        self.filter_all_unknown_hosts = False
         self.filter_all_flapping_hosts = False
         self.filter_all_unknown_services = False
         self.filter_all_information_services = False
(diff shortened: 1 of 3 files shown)


HenriWahl:
@blaboucheShinken thanks - now I fully understood the intention. Yes, only services were able to be UNKNOWN. 👍️

blaboucheShinken:
@HenriWahl you're welcome, thank you for merging and for your work, I am closing the issue :)

HenriWahl:
It is now in the latest binary, despite the version number did not change yet.

Issue #1204, opened 2026-03-26
Livestatus servers are not able to connect with ssl


blaboucheShinken:
Hello, 

We noticed that when using Livestatus servers it is not possible to establish a connection using ssl protocol.
We changed the way Livestatus servers initialize sockets to allow using ssl. A Pull Request is coming soon.

Best regards,
Bastien from Shinken Solutions

HenriWahl:
Merged already.

PR #1205, opened 2026-03-26
Adding ssl support for Livestatus servers

blaboucheShinken:
Hello, 

We noticed that when using Livestatus servers it is not possible to establish a connection using ssl protocol.
This is the Pull Request related to the #1204.
We now collect the protocol from the self.monitor_url property and use it to determine if we should use ssl or not, then we initialize the socket with a contextmanager, depending on the protocol and the options defined in server configuration (like the ignore_cert option).
We also changed the "communicate" function to use the "with" statement to automatically close the sockets.

We are open to any comment, if you feel that some changes need to be made or if you would like things to be handled differently, feel free to let us know and we will adjust accordingly.
Best regards,
Bastien from Shinken Solutions


HenriWahl:
@blaboucheShinken thank you, it's now in latest testing release.

PR #1218, opened 2026-04-30
Fix QTreeView item hover shift and excessive row padding

### LINKED ISSUE #1217: QTreeView item stylesheet: row shift on hover and excessive padding
The stylesheet applied to `QTreeView::item` in `treeview.py` has two related problems:

1. **Row content shifts on hover.** Normal items use `margin: 5px` while hover/selected states switch to `margin: 0px; padding: 5px`. These are different box model properties and cause the row content to visibly jump a few pixels to the right when the cursor enters the row.

2. **Excessive row padding.** The `margin: 5px` on the base item state makes rows taller than necessary.

**Fix:** Use `padding` consistently across all item states (base, `:hover`, `:selected`) and suppress the focus rectangle with `QTreeView { outline: 0 }`. This eliminates the shift, reduces row height, and keeps the hover/selection highlight flush across cells.

### LINKED ISSUE #1219: QTreeView item hover/selection gives no visual feedback without overriding status colours
The `:hover` and `:selected` stylesheet rules replace the row's status background colour (CRITICAL red, WARNING yellow, etc.) with a static `dimgrey`/`grey`, which is unintuitive — the alert context is lost on hover/selection.

A better approach is to leave the status background intact and use the system highlight colour (`palette(highlight)`) as top/bottom borders to indicate hover and selection. This preserves the status colour at all times and works correctly in both dark and light themes.

codemedic:
## Summary

- Replace the mixed `margin`/`padding` stylesheet on `QTreeView::item` with consistent `padding: 3px` across all states (base, `:hover`, `:selected`)
- Add `QTreeView { outline: 0 }` to suppress Qt's focus rectangle on hovered/selected rows
- Replace static `dimgrey`/`grey` hover/selection backgrounds with `palette(highlight)` top/bottom borders, preserving the row's alert status colour

## Details

The original stylesheet used `margin: 5px` on the base item state but switched to `margin: 0px; padding: 5px` on hover and selected states. Because `margin` and `padding` are different box model properties, this caused row content to visibly jump a few pixels sideways when hovering. Using `padding` consistently eliminates the shift and keeps the hover/selection highlight flush across all cells.

The static `dimgrey`/`grey` background on hover/selection replaced the row's alert status colour (CRITICAL red, WARNING yellow, etc.), losing the visual context. Using `palette(highlight)` borders instead keeps the status colour visible at all times and adapts correctly to both dark and light themes.

Closes #1217.
Closes #1219.

### DIFF
--- a/Nagstamon/qui/qt.py
+++ b/Nagstamon/qui/qt.py
@@ -142,6 +142,7 @@ if QT_FLAVOR == 'PyQt5':
     from PyQt5.QtWidgets import QAbstractItemView, \
         QAction, \
         QApplication, \
+        QStyledItemDelegate, \
         QColorDialog, \
         QComboBox, \
         QDialog, \
@@ -283,6 +284,7 @@ elif QT_FLAVOR == 'PyQt6':
     from PyQt6.QtWebEngineWidgets import QWebEngineView as WebEngineView
     from PyQt6.QtWidgets import (QAbstractItemView,
                                  QApplication,
+                                 QStyledItemDelegate,
                                  QColorDialog,
                                  QComboBox,
                                  QDialog,
--- a/Nagstamon/qui/widgets/treeview.py
+++ b/Nagstamon/qui/widgets/treeview.py
@@ -73,13 +73,26 @@ class StatusAwareDelegate(QStyledItemDelegate):
     """
     Renders hover and selection as reverse video using each row's own
     status colours, so the alert context is always visible.
+
+    Drawing order: super().paint() first (handles icons, decorations, font),
+    then we overdraw background and text. This wins over native OS styles
+    (Windows/macOS) that ignore option.backgroundBrush inside drawControl.
     """
 
-    def initStyleOption(self, option, index):
-        super().initStyleOption(option, index)
+    # QPalette colour role access differs between PyQt5 (<5.15) and PyQt6
+    try:
+        _ROLE_TEXT = QPalette.ColorRole.Text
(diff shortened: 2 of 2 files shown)


HenriWahl:
Hi @codemedic I merged into an extra branch https://github.com/HenriWahl/Nagstamon/tree/qtreeview_improvements for checking it out.
I run it under Linux and the first thing I find a little buggy is, that a highlit row switches back to its original colors when the status info gets reloaded. Maybe the delegation is causing this? Maybe there is a way to use the system highlight color? I remember there was some problem I run into that's why I took that neutral grey thing.
The nice thing I see that the indent does not jump anymore under Linux.

codemedic:
Hi @HenriWahl — thanks for reviewing!

You're right about the highlighted row reverting on reload — that's a known issue with the delegate approach: Qt clears `State_MouseOver` during model resets and doesn't re-fire hover events until the mouse moves. A fix is in progress on [fix/treeview-hover-flicker-on-refresh](https://github.com/codemedic/Nagstamon/tree/fix/treeview-hover-flicker-on-refresh) and will follow as a PR against `qtreeview_improvements` shortly.

On the system highlight colour — I did consider that route, but found it unreliable across platforms and themes. On Ubuntu with the default dark theme, the system highlight is orange, which clashes badly with Nagstamon's status colours. On Windows dark mode it was equally problematic. The system highlight colour is driven by OS and desktop environment preferences, which are outside Nagstamon's control and can conflict with any of the alert colours.

Reverse video sidesteps this entirely — it always uses the row's own status colours inverted, so the alert context is always clear regardless of OS or desktop theme.

PR #1222, opened 2026-05-03
Fix hover highlight lost on status refresh

codemedic:
## Summary

- Tracks the hovered row via the view's `entered` signal rather than relying solely on `State_MouseOver` in `initStyleOption`
- `State_MouseOver` is transiently absent during model-reset repaints (triggered on every status refresh), causing the hovered row to lose its reverse-video highlight until the mouse moves
- A viewport event filter clears the tracked row when the mouse leaves, preventing stale highlighting

## Test plan

- [x] Hover over a row and wait for a status refresh — highlight should remain stable throughout
- [x] Move the mouse off the treeview — highlight should clear immediately
- [x] Move the mouse quickly across rows — no multi-row or partial-row artefacts
- [x] Verify clean exit (no crash on application quit)
- [x] Test on Linux, Windows, and macOS

Fixes the regression reported in #1218.

### DIFF
--- a/Nagstamon/qui/qt.py
+++ b/Nagstamon/qui/qt.py
@@ -106,6 +106,7 @@ if QT_FLAVOR == 'PyQt5':
         QAbstractTableModel, \
         QByteArray, \
         QDateTime, \
+        QEvent, \
         QModelIndex, \
         QObject, \
         QPoint, \
@@ -248,6 +249,7 @@ elif QT_FLAVOR == 'PyQt6':
                               QAbstractTableModel,
                               QByteArray,
                               QDateTime,
+                              QEvent,
                               QModelIndex,
                               QObject,
                               QPoint,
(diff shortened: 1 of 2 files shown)


codemedic:
@HenriWahl does that look better?

HenriWahl:
@codemedic definitely! Cool, thanks!

HenriWahl:
Hi @codemedic - I am running your improved version on a daily base now and am really fast getting used to the design change 😀. Just one note: When using the context menu and starting a recheck of a single service, the highlighting stays there. Might be a Qt event loop problem, where the Treeview worker interferes somehow. Maybe I need to put the background recheck jobs into some extra thread anyway.

codemedic:
Hi @HenriWahl — thanks so much for the kind words, really glad it's working out day-to-day!

Good catch on the Recheck case. Looking into it now — turns out it's a different mechanism than the model-reset hover issue, so a separate follow-up. I'll send the fix along shortly.

codemedic:
@HenriWahl I think I have a fix in https://github.com/HenriWahl/Nagstamon/pull/1226

Issue #1225, opened 2026-05-12
Entra SSO - Alertmanager OK not Alerting


lunachocken:
## Description
SSO protected endpoint via Entra for alertmanager does not trigger authentication and instead returns OK.

## Reproducing setup
- Envoy Gateway on ingress
- Httproute for grafana.xxx.yyy.zzz with path /alertmanager (/alertmanager redirects to alertmanager pod not Grafana)
- OIDC security policy on httproute with redirect for SSO
- Organisation Microsoft Entra as the SSO issuer

## Steps to reproduce
1. Set monitor type to "Alertmanager"
1. Set authentication to "web"
2. Set url to https://grafana.xxx.yyy.zzz/alertmanager
3. Confirm settings
4. Observe "OK"

## Expected behaviour
1. Nagstamon triggers popup requesting authentication
2. Login to https://login.microsoftonline.com/ 
3. Close window
4. Several alerts appear

## Actual behaviour
1. Upon configuring and setting configs. A simple "OK" is present, no request for authentication, and doesn't actually get past the SSO.

## Environment
OS:
```
NAME="Nobara Linux"
VERSION="43 (KDE Plasma Desktop Edition)"
```
Environment: distrobox with image: fedora-toolbox
- Nagstamon:  nagstamon.noarch 3.19.20260508-1 nagstamon-latest

## Reflection

I assume the error is due to microsoft's page returning a 200 status, which nagstamon assumes is fine. 
On using mitmproxy to change the status code to 401 instead, this triggers nagstamon to require authentication and resolves the issue.

```
from mitmproxy import http

def response(flow: http.HTTPFlow):
    # 1. Check if the host matches Microsoft Login
    if "login.microsoftonline.com" in flow.request.pretty_host:

        # 2. Check if the path contains /oauth
        if "<tenant-id>/oauth2/v2.0/authorize?" in flow.request.path:

            # 3. Change 200 OK to 401 Unauthorized
            if flow.response.status_code == 200:
                flow.response.status_code = 401
                #flow.response.reason = "Unauthorized"
                #flow.response.text = "Access Denied: Microsoft Login Intercepted"

                print(f"DEBUG: Intercepted {flow.request.url} - Forced 401")
```

PR #1227, opened 2026-05-22
fix: ms login page not triggering authentication

lunachocken:
This resolves my issue at: #1225 

When the response returns a Microsoft login page with a 200 status code, the status code is instead returned as 401 mandating the user to login.

### DIFF
--- a/Nagstamon/servers/Generic.py
+++ b/Nagstamon/servers/Generic.py
@@ -1108,7 +1108,7 @@ class GenericServer:
                         if conf.debug_mode:
                             self.debug(server=self.get_name(), debug='Filter: UNKNOWN ' + str(host.name))
                         host.visible = False
-    
+
                     if host.visible:
                         self.nagitems_filtered['hosts']['UNKNOWN'].append(host)
                         self.unknown += 1
@@ -1600,7 +1600,12 @@ class GenericServer:
                 self.encoding = response.encoding
 
             # Microsoft Authentication required - Required since mslogin returns 200
-            if response.status_code == 200 and "<!DOCTYPE html>" in response.text and "https://login.microsoftonline.com/" in response.text and self.authentication == 'web':
+            if (
+                self.authentication == "web"
+                and response.status_code == 200
+                and "<!-- Copyright (C) Microsoft Corporation" in response.text[0:100]
+                and "https://login.microsoftonline.com/" in response.text[0:2048]
+            ):
                 response.status_code = 401
 
             # give back pure HTML or XML in case giveback is 'raw'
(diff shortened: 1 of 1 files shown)


HenriWahl:
Hi @lunachocken - nice workaround, but what do you think of a solution, where this fix won't be checked with every web request ever done?

lunachocken:
Hmm.

I suppose it might be possible to use a cache of some sort. A function that sets the status code if it's detected to be a login page.

Then set a cache time before the checks run again. 

The alternative, though not sure how to reuse code in this case. Would be to check it on every healthcheck of the server, if there is one and just add to that.

Though I shouldn't think it'd be a huge overhead in the first place?

HenriWahl:
Well, you are right, this sounds overcomplicated as well. How about changing the order of the conditions to avoid further processing, like from

```
if response.status_code == 200 and "<!DOCTYPE html>" in response.text and "https://login.microsoftonline.com/" in response.text and self.authentication == 'web':
```
to
```
if response.text.startswith( '<!DOCTYPE html>') and \
   response.status_code == 200 and 'https://login.microsoftonline.com/' in response.text and \
   self.authentication == 'web':
```
Then at least all the JSON/XML based answers of certain monitoring servers will not be scanned further.
Can you check if this would work too?

lunachocken:
I was having a look at python on if statement ordering, I'm not too well versed in it. 

Since it's evaluated left-to-right, the web and status_code should be the fastest. Since I assume people will not predominantly be using web auth, this immediately skips the other checks.

```python
if self.authentication == 'web' and response.status_code == 200 and \
   response.text.startswith( '<!DOCTYPE html>') and \
   'https://login.microsoftonline.com/' in response.text:
```

Just a minor thing, move the string matching to after the basic variable checks.

Though the startswith is failing the script. A response is:
````
2026-05-27 16:02:54,718 - alertmanager - DEBUG - received status code '200' with this content in result.result:
                           ---------------------------------------------------------------


<!-- Copyright (C) Microsoft Corporation. All rights reserved. -->
<!DOCTYPE html>
<html dir="ltr" class="" lang="en">
``

**Comment by HenriWahl** [comment]
Nice finding or at least nice that you found this behavior of the MS login service. Well, syntactically correct but surely no standard... so you are right, the `.startswith()` is useless here.

**Comment by HenriWahl** [comment]
I merged into https://github.com/HenriWahl/Nagstamon/tree/mslogin_workaround for further exploration. One more idea: what about not looking for 'https://login.microsoftonline.com' in `response.text` but `response.url`? I can't test it due to not having an Entra account but could you check please? Maybe even looking for the Copyright nonsense would be obsolete, just checking for `response.status_code == 200` and `response.url.startswith('https://login.microsoftonline.com/')`?

**Comment by lunachocken** [comment]
```python
            if (
                self.authentication == "web"
                and response.status_code == 200
                and response.url.startswith("https://login.microsoftonline.com")
            ):
                response.status_code = 401
```

Seems to work just fine

Issue #491, opened 2018-05-07
filter host group


swizzly:
Hi 
Is it possible to make a filter for host groups and not only hostnames, e.g. I would like to display only hosts that are in the host group "Windows" in Icinga2.
Cheers

swizzly:
Ok, I found a Workaround. Using the extragroups module for Icingaweb2 custom groups can be assigned to the user using a given web-agent e.g. nagstamon. So for example I want to display a given hostgroup in nagstamon I can define the hostgroup in Icinga2 and then configure in the extragroups module to show me only that given hostgroup if I use the nagstamon agent.
It works nicely, but of course an out of box solution from the nagstamon side would be more elegant.

donestefan:
There is some unfinished code to add this feature in #577. But the issue was closed due to inactivity 😥

PR #1231, opened 2026-06-04
[#491] Add host group filtering for IcingaWeb2 and Icinga2API

rvallo:
- IcingaWeb2: fetch host groups via separate per-hostgroup requests, propagate groups to services for service-level filtering
- Icinga2API: populate host.groups from API response
- Settings: show groups filter widget for IcingaWeb2 and Icinga2API

Assisted-by: Claude:sonnet-4-6

(Resolving #491)

### DIFF
--- a/Nagstamon/qui/dialogs/settings.py
+++ b/Nagstamon/qui/dialogs/settings.py
@@ -355,6 +355,11 @@ class DialogSettings(Dialog):
                                    self.window.input_lineedit_re_groups_pattern,
                                    self.window.input_checkbox_re_groups_reverse]
 
+        # same groups filter widgets, additionally shown for Icinga server types (issue #491)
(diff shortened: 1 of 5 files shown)


Nagstamon/qui/dialogs/settings.py
```diff
@@ -1025,15 +1025,15 @@ def toggle_zabbix_widgets(self):
     @Slot()
     def toggle_op5monitor_widgets(self):
```
HenriWahl, line 1032: Please do not mix with Op5Monitor to avoid later confusion - shouldn't this become some extra toggle?

rvallo: OK, I rewrote it :)

rvallo:
I found some bugs, I'll rewrite it and create a PR later.

Issue #1235, opened 2026-07-06
Nagstamon not working with latest version of Icingaweb2


adripa45:
Hi,

Nagstamon (3.18.2) seems broken since I updated Icingaweb2 to the latest version : 2.14.0

I use the IcingaDBWeb type and had no problem until the update.

Here's some log : 

```
DEBUG: 2026-07-06 15:56:21.872062 Monitor_IcingaMaster-CDS fetch_url: https://myicingaserver/authentication/login CGI Data: None
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/Nagstamon/qui/widgets/treeview.py", line 1038, in get_status
    status = self.server.get_status()
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3/dist-packages/Nagstamon/servers/Generic.py", line 951, in get_status
    self.init_http()
  File "/usr/lib/python3/dist-packages/Nagstamon/servers/IcingaDBWeb.py", line 133, in init_http
    form_inputs[form_input] = form.find('input', {'name': form_input})['value']
                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
  File "/usr/lib/python3/dist-packages/bs4/element.py", line 1573, in __getitem__
    return self.attrs[key]
           ~~~~~~~~~~^^^^^
KeyError: 'value'
```

Best regards

benjamingufler:
Same issue here. Login to icingaweb2 server doesn't work with Icingaweb 2.14.0. Instead, nagstamon crashes.

oliver:
At #1236 there is an attempt to fix this. So far it works well for me.

If someone has access to an older Icingaweb2 instance (before 2.14), it would be very useful to test the PR against that version, to make sure the fix did not break compatibility with older Icingaweb2!

kuly2fraise:
Hello @oliver ,

I have an old Icingaweb2 running on version 2.13, I can test if you just tell me how to compile your updated version please. I'm using the windows one

Let me know, I will be happy to help :)

kuly2fraise:
Hello again @oliver ,

I was able to test it with an icinga host (version 2.13). Thank you for your work!

In my specific case, I also had to apply the same change in the IcingaDBWeb part.

The required modification is in /Nagstamon/servers/IcingaWeb2.py:
`        # normally cookie auth will be used
        if not self.no_cookie_auth:
[...]

PR #1236, opened 2026-07-06
fix login with Icingaweb2 2.14

oliver:
This fixes login to Icingaweb2 version 2.14, by adapting to the changed form fields on login page. It should fix issue #1235 .

Unfortunately I was not able to test with older Icingaweb2 versions (before 2.14). The new code might still work with those versions, but I'm not sure.


HenriWahl:
@oliver thank you - I am on vacation now, but will check your PR as soon as back!

alexanderclarktx (approved):
Approved on Oliver's request.

PR #1238, opened 2026-07-07
Fix : IcingaDBWeb causing crashes to nagstamon

kuly2fraise:
This pull request updates the login form handling in the `init_http` method of `IcingaDBWeb.py` to improve compatibility with additional form fields and handle missing values more robustly.

This fixes the crash issue when trying to access an Icinga host using the IcingaDB Web module.

The fix targets Icinga Web 2 version 2.14 and higher, where the issue was introduced. Older versions remain compatible and are still working.

**Login form parsing improvements:**

* The list of expected form input fields has been expanded to include `uid` and `submit_login`, in addition to the existing fields.
* The code now checks for the presence of each input field and its `value` attribute before assigning it, defaulting to an empty string if not found. This makes the parsing more robust against missing or incomplete form fields.

Ref issue: #1235

Happy to help.

### DIFF
--- a/Nagstamon/servers/IcingaDBWeb.py
+++ b/Nagstamon/servers/IcingaDBWeb.py
@@ -128,11 +128,13 @@ class IcingaDBWebServer(GenericServer):
                 if login.error == '' and login.status_code == 200:
                     form = login.result.find('form')
                     form_inputs = {}
-                    for form_input in ('redirect', 'formUID', 'CSRFToken', 'btn_submit'):
-                        if form is not None and not form.find('input', {'name': form_input}) is None:
-                            form_inputs[form_input] = form.find('input', {'name': form_input})['value']
-                        else:
-                            form_inputs[form_input] = ''
+                    for form_input in ('redirect', 'formUID', 'CSRFToken', 'btn_submit', 'uid', 'submit_login'):
+                        form_field_value = ''
+                        if form is not None:
+                            form_field = form.find('input', {'name': form_input})
+                            if form_field is not None and form_field.has_attr('value'):
+                                form_field_value = form_field['value']
+                        form_inputs[form_input] = form_field_value
                     form_inputs['username'] = self.username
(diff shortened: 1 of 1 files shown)


kuly2fraise:
Based on #1236 solution. 

Thank you @oliver :)

HenriWahl:
@kuly2fraise thank you - I am on vacation now, but will check your PR as soon as back!

metrax:
@HenriWahl  are you still on vacation? A lot of user needs this fix as soon as possible.

HenriWahl:
> @HenriWahl are you still on vacation? A lot of user needs this fix as soon as possible.

Yes, to sad I am not anymore... well, I built a new testing release with the merge (thanks!). Please note it is unsigned for Windows as the signing key expired and I am about to test a new solution based on a hardware stick.

chas0rde:
> > @HenriWahl are you still on vacation? A lot of user needs this fix as soon as possible.
> 
> Yes, to sad I am not anymore... well, I built a new testing release with the merge (thanks!). Please note it is unsigned for Windows as the signing key expired and I am about to test a new solution based on a hardware stick.

Do you have any idea when a signed version might be available?
I cannot use an unsigned one and currently am running half blind without nagstamon ;)

HenriWahl:
I at least managed to sign with the new hardware key, which is now required. I fear this cannot be integrated sensefully into every CI pipeline, so maybe only stable releases may come signed. I hope I will find a clever way as compromise, maybe sign somehow delayed, when the hardware key is available.

HenriWahl:
@chas0rde can you at least prove that the last testing release works with Icinga 2.14, even if unsigned?

HenriWahl:
@chas0rde I finally managend to sign the binaries - please check the latest testing release.

chas0rde:
Hello
sorry no cannot test the unsigned one as I am not allowed to run it on our systems due to security precautions
Where do I find the testing releases?

Issue #1148, opened 2025-10-19
When auto starts up from login, does not set sizing correctly


Outsidewall:
Hello, When nagstamon automatically starts up from login, it does position it's self where it was last, however it does not set it's sizes correctly (width, height)

HenriWahl:
Hey @Outsidewall - what OS on do you use which version of Nagstamon?

Outsidewall:
Windows 11

HenriWahl:
@Outsidewall which Nagstamon version are you using?

PR #1243, opened 2026-09-01
macOS: hint when running under Rosetta, and a start at login option

discostur:
> Builds on #1241 - GitHub shows its three commits here as well until that one is merged. Only the last two commits belong to this PR.

Two small macOS things.

### Tell the user when Nagstamon runs under Rosetta

The Qt binaries are not available as universal binaries, so there is one DMG per architecture. Downloading the Intel one on an Apple Silicon Mac works, macOS just runs it under Rosetta, which makes it a quiet trap: the crash report in #1055 shows `Code Type: X86-64 (Translated)` on a Mac that should have used the ARM build, and the wrong download was never suspected in that thread.

Nagstamon reads `sysctl.proc_translated` at startup now, prints a hint and puts it into the debug output, and the About dialog shows the running architecture next to the Python and Qt versions.

### Start at login

Windows gets its autostart entry from the installer and Linux from the `.desktop` file, but macOS users have to add Nagstamon to their login items by hand. The settings offer a "Start at login" checkbox now which writes a LaunchAgent to `~/Library/LaunchAgents`.

A LaunchAgent rather than `SMAppService` because it needs no additional PyObjC dependency and is a single file that can be removed again without a trace. It still shows up in the login items in the system settings. The configuration directory is passed on, so an instance started at login uses the same configuration as the one the setting was made in.

The checkbox only appears when Nagstamon runs from an application bundle - there is nothing sensible to point a login item at otherwise - and its state is read back from the LaunchAgent, because the user may have removed the login item in the system settings in the meantime.

This also touches the reason a login item was awkward in the first place. When Nagstamon starts at login the screens are not necessarily enumerated yet, so the stored window position looks off-screen, gets discarded in `set_mode()` and the window comes up with a default position and size. The status window remembers now whether it could use its stored position and re-applies it once `QApplication` reports a new or changed screen. That should be #1148, though I could not reproduce the reporter's exact setup.

Tested on macOS 26.6.2 on Apple Silicon.


HenriWahl:
Hey @discostur - I do not run Nagstamon on a Mac, so I cannot test this fully. I see some parts regarding the hiding of the icon at the dock are vanished now in the code - how does it behave now? Are the dock icons still invisible?

discostur:
Hi @HenriWahl - nothing of the dock icon hiding is gone, and I tested it locally on macOS (from source, PyQt6 6.11) to be sure.

The code paths which hide the icon are untouched: `StatusWindow.set_mode()` still calls `hide_macos_dock_icon(conf.hide_macos_dock_icon)` in systray mode (and `hide_macos_dock_icon(False)` for floating and fullscreen), and the dialogs still show the icon while a dialog is open and hide it again afterwards.

What the PR removes are three pieces which did nothing:

- `Nagstamon/qui/__init__.py` imported `NSApp`, `NSApplicationPresentationDefault` and `NSApplicationPresentationHideDock`, but no line in that module ever used them - the implementation lives in `Nagstamon/qui/helpers.py`, where the imports are still there.
- `Nagstamon/qui/dialogs/__init__.py` connected `dialogs.authentication.check_macos_dock_icon_fix_show`/`_hide` a second time to the same two slots, two lines below the first pair.
- `statuswindow.py` had the `# test with OSX top menubar` block: `self.menubar` was built with two `QAction`s, but it was never attached to a window and the actions were never connected - `menubar` is not referenced anywhere else in the tree.

The one real behaviour change is the constants. `hide_macos_dock_icon(True)` passed `NSApplicationPresentationHideDock` to `setActivationPolicy_()`. That is a presentation-options constant which happens to have the value 2 - and 2 as an activation policy is `NSApplicationActivationPolicyProhibited`. Prohibited hides the dock icon, but it also keeps the application from being activated at all, which is why windows could not take keyboard focus. Measured here:

```
NSApplicationPresentationHideDock      = 2
NSApplicationActivationPolicyProhibited= 2   # what the old code really set
NSApplicationActivationPolicyAccessory = 1   # what it sets now
NSApplicationActivationPolicyRegular   = 0

Prohibited (old code)  app_active=False key_window=False focus_widget=False
Accessory (new code)   app_active=True  key_window=True  focus_widget=True
Regular (icon shown)   app_active=True  key_window=True  focus_widget=True
```

And with a running Nagstamon with `icon_in_systray` plus `hide_macos_dock_icon` enabled:

```
this branch:  activationPolicy=1 (Accessory)   lsappinfo: ApplicationType="UIElement"
master:       activationPolicy=2 (Prohibited)
```

`ApplicationType="UIElement"` is macOS saying the process has no dock tile, so the icon is still invisible - it is just usable now.

While I had it running I also checked the quit path of #1241 with a server pointing at a blackhole address, so a status request was hanging in `connect()` when the quit arrived (three threads stuck in the sample):

```
this branch:  quit finished after 2.1s, no "QThread: Destroyed while thread is still running", no crash report
master:       "QThread: Destroyed while thread '' is still running"
```

Happy to test anything else you would like to see before merging.

